Issue #20, opened 2021-11-15
Add recording links to Brave Talk


mrose17:
How things work right now:
- when a recording starts/stops, all participants receive an audio warning (e.g., "Recording is on"). In addition, the red REC indicator is visible at the top of the window whenever recording is occurring.
- when the moderator starts a recording, the moderator gets a notification with a "Copy Link" to allow the moderator to down the link a few minutes after the recording has been stopped by the moderator (or all attendees have left the room).
- recordings are no longer available 24h after the recording ends.

API changes:
- the `recordingStatusChange` event is sent to all participants when a recording/stops starts, i.e., `{ "on": true, "mode": "file" }` or `{ "on": false, "mode": "file" }`
- the `recordingLinkAvailable` event is sent only to the moderator when recording starts, e.g., `{ link: "https://..." }`
- a new URL -- `https://api-vo.jitsi.net/jaas-recordings/link/details?url={{URL}}`--  is available to return `{ expiresAt: ... }`

Site changes (for moderator only):
- the recording URL along with room name, creation and expiration timestamps are stored in `localStorage`, and entries are automatically pruned as they expire.
- when the site loads, this information is displayed to the user allowing them to download the files.
- the new URL is not used in this implementation.

here is a sample from the development site. (note that the styling isn't done...)
<img width="1532" alt="Screen Shot 2021-11-22 at 12 19 25" src="https://user-images.githubusercontent.com/400552/142925324-3c6a885d-6cd9-4798-8068-16c7a4b54e80.png">
-

tackley:
Release to production in #110 

![Screenshot 2022-01-03 at 16 35 17](https://user-images.githubusercontent.com/51444/147956998-3ee6be98-33b9-4178-a9b4-5c0326e0e227.png)

PR #50, opened 2021-11-25
Recording updates

mrose17:
cf., https://github.com/brave/brave-talk/issues/20


src/store.ts
```diff
@@ -13,13 +13,33 @@
 
 export interface JwtStore {
```
diracdeltas, line 14: should this just be renamed to Store since as i understand it, some of these methods have nothing to do with JWTs?

src/store.ts
```diff
@@ -29,15 +49,57 @@ export function loadLocalJwtStore(): JwtStore {
...
+          ttl +
+          "  createP=" +
+          !confabs.recordings[url]
+      );
+
+      const now = Math.ceil(new Date().getTime() / 1000);
+      if (typeof ttl === "undefined") {
+        ttl = 24 * 60 * 60;
+      }
+      const expiresAt = now + ttl;
```
diracdeltas, line 89: does the server calculate expiration separately from this? otherwise a client who wants their recording to persist for 24+ hours can set `expiresAt` to whatever they want

mrose17: The 24h is hard-wired into the 8x8 back-end. i have asked them to add a field in the API response to indicate the TTL, so if/when it changes in the future, reality will be enforced...

src/store.ts
mrose17: @tackley had the same idea!

graham - what's a good new name?

diracdeltas:
this looks reasonable to me but this repo needs `npm audit` set up to run as a github action for every PR. can be done as part of this one or separately. also should have a test suite run in github actions as well.

src/store.ts
tackley: after cycling round on this, we've decided to keep the store of recordings separate to the jwt one. They're quite different things.

tackley (approved):
It's been a joint journey, but this looks great to me...

Issue #16, opened 2021-11-11
give `/talk` non-subscribers options to create the call anyway and to login


tackley:
brave/brave-together#258 implemented logic to try to resolve the subscriber vs non-subscriber race condition when starting a call from the slack `/talk` plugin. This works great if at least one person trying to join the call is already logged in as a subscriber on the browser that slack chose to open.

However, if that's not the case, it appears to get stuck in a loop and the user may have to resort to using a competitors' product. The only way out of the loop is to hack the url to remove `?create=y`.

See e.g. https://bravesoftware.slack.com/archives/C01BC0K5QP7/p1636646236160700 and https://bravesoftware.slack.com/archives/C01BC0K5QP7/p1636390882144400.

 When showing the "waiting for a subscriber" message we should additionally provide buttons to "create a 1:1 room" and "login".

tackley:
Now live in production

PR #57, opened 2021-11-30
give `/talk` non-subscribers options to create the call anyway and to login

tackley:
This is not totally ideal as when you login, it loses the room name you were trying to connect to, so you need to go back to slack and click the link again :(  But it's better than getting stuck, and at least you can start a 1:1 room.

Re #16

https://user-images.githubusercontent.com/51444/144041013-9fb5c5cf-2e5a-4cea-ad91-1602579705d9.mov


mrose17 (approved):
LGTM

PR #69, opened 2021-12-02
Production Release

### LINKED ISSUE #16: give `/talk` non-subscribers options to create the call anyway and to login

tackley:
- closes #16 
- build time dependency version bumps


mrose17 (approved):
Winner!

PR #70, opened 2021-12-02
Update the skus-sdk to the publicly released version

tackley:
And while I was in the area:
- make the logging in subscriptions a bit less repetative
- remove the config of a private access token in the github actions (not needed as @brave-intl/skus-sdk is public)
- improve consistency between the github actions


mrose17 (approved):
Winner!

PR #71, opened 2021-12-02
Production release - upgrade rewards sdk

tackley:
- #70

PR #72, opened 2021-12-03
Check "active" value on skus-sdk

tackley:
See https://github.com/brave/brave-talk/pull/71#discussion_r761326423


mrose17 (approved):
That works!

PR #75, opened 2021-12-06
Fetch credentials if order exists but has no credentials

tackley:
This logic was there previously, but failed because it didn't take into account the changed contract for `credential_summary` in the new sdk (where it now returns a value with `active` set to false in this case) . Removed the duplication of this check so it should not happen again.

Net effect was that new logins did not work as the credentials were not fetched, see [slack](https://bravesoftware.slack.com/archives/C021S0FN3V1/p1638739925331200).


mrose17 (approved):
LGTM

PR #80, opened 2021-12-09
Support `create_only=y` parameter

tackley:
To create a room then close the window, used by the gcal plugin.

TODO: display better messaging if they are  not a subscriber, rather than just creating a 1:1 room.


mrose17 (approved):
LGTM

PR #86, opened 2021-12-13
Use style-loader in dev mode

tackley:
Which means that css changes hot-reload. See https://github.com/webpack-contrib/mini-css-extract-plugin#recommend.

Has no effect on the production release, but makes making css changes locally far easier.


mrose17 (approved):
LGTM

Issue #81, opened 2021-12-09
Refresh tokens disappearing


mrose17:
Somehow the refresh token for one of my rooms disappeared? How, I don't know. All I know is I went to the room and the JWT didn't have moderator-privileges. I am sure that I was in the room within the last 10 days!

tackley:
In both reported cases, the room was no longer present within the `refresh` element of the `confabs` localStorage entry. @mrose17 I believe you reported that your refresh element was completely empty, correct?

There are three different conditions that I can think of might cause this:
1. The localStorage entry that contains jwts and refresh tokens got accidentally cleared, either by a code within brave-talk or some other interaction. It doesn't look like localStorage was completely reset, as both reporters were still logged in meaning that the `skus:production` local storage entry was still present.
2. We incorrectly [garbage collected](https://github.com/brave/brave-talk/blob/cd245c776169b1186f6bba36de9a744155fde108/src/jwt-store.ts#L101-L106) the refresh tokens before they actually expired.
3. A race condition occurs on reading / writing the JWT store which means that in some cases this data gets lots. This doesn't look likely as the store is [read](https://github.com/brave/brave-talk/blob/main/src/rooms.ts#L165) only within the `fetchJWT` method - however this method has may async calls within it, so this might be possible.

Without access to historical browser logs, it's difficult to diagnose further. However, to aid diagnosis in the future:
a. temporarily stop garbage collecting the refresh tokens so we can still inspect expired ones if this issue recurs.
b. change the logic of jwt store so it loads once on first request, and from that point on the in-memory copy is considered the master.

mrose17:
just to clarify: I did have refresh token for two other rooms that i created...

mrose17:
i agree with the two changes (the first one being temporary until we figure out what's going on).

mrose17:
unable to reproduce. _sigh_ (i think).

mrose17:
Unable to reproduce, cf., #184

PR #87, opened 2021-12-13
Refresh tokens disappearing

tackley:
Don't think this is a fix, but should aid diagnosis:
- treat the in-memory copy of "confabs" as the master after initial load
- don't expire refresh tokens for now
- add a unit test
Re #81


mrose17 (approved):
LGTM

PR #100, opened 2021-12-16
Disable before unload handlers

mrose17:
Fixes #8

PR #101, opened 2021-12-16
Production Release 2021-12-16

tackley:
- #97 (device local circular log for diagnosing refresh token issues) 
- #95 
- #100

NB: https://github.com/brave-intl/subscriptions/pull/263 should be deployed **before this** so that the promise of 4 person rooms is delivered upon.


mrose17 (approved):
LGTM

tackley:
https://github.com/brave-intl/subscriptions/pull/263 is live in production, so this is ready to go.

Issue #104, opened 2021-12-16
Needs automated tests + npm audit + codeql checks running on every PR


diracdeltas:
See https://github.com/brave/brave-talk/pull/50#issuecomment-996230384 for details. codeql is set up in https://github.com/brave/brave-talk/pull/103 but may need someone to set it up as a required status check for PRs in github settings after it is merged.

diracdeltas:
It looks like there are some tests which get run on deploy, but it's unclear to me if these are run on every PR and will block the PR from being merged if so. https://github.com/brave/brave-talk/runs/4551728270?check_suite_focus=true

tackley:
Thanks for the spot here @diracdeltas - we hadn't noticed this as our working process is never to merge or approve a PR unless we've seen the code running on the development environment, the deploy for which (like all of the deploys) require tests / audit etc to pass before deploying.

However that's not necessarily clear to people outside the project team, and our github actions needed a bit of love to reduce duplication anyway. So #108 addresses this. Also updated the branch protections to require status checks on `main` and `prod`.

diracdeltas:
thanks @tackley !

PR #108, opened 2021-12-17
CI Improvements

tackley:
- reduce the copy-paste between scripts (though sadly the verbosity of composite actions mean this doesn't actually reduce file size at all - the opposite in fact - but the commonality of script is worth it I think)
- run build, unit-test and audit on every push

Re #104


mrose17 (approved):
Excellent, thank you!

PR #138, opened 2022-01-13
Dependabot interval change

mrose17:


tackley (approved):
As discussed, this make sense to avoid some of the endless chasing of dependabot PRs. Security-releated ones come via a different route and will still apply immediately.

PR #139, opened 2022-01-13
Remove email field and get current with dependabot...

mrose17:


tackley (approved):
Looks great, this is the set of changes that we've previously reviewed when merging to `main`.

- #129 
- #136 
- #139

PR #140, opened 2022-01-16
WHY DID YOU HAVE SO MANY EXTRA DIVS

WatchGamingVideos:
Pls approve

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -207,11 +207,5 @@
                   </div>
                   <div class="opt-in-step-2-screenshot"></div>
                 </div>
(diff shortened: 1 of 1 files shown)


mrose17:
@WatchGamingVideos - i don't understand this PR. the HTML has balanced `<div>` and `</div>` tags. as far as I can tell, everything is nested properly. the PR removes the last six `</div>` tags but not the corresponding `<div>` tags, so applying the PR would "break" the HTML.

mrose17:
Closed due to lack of response.

PR #160, opened 2022-01-25
Fully clear cloudfront cache on deploy

tackley:
I'd tweaked this to just clear the index.html. However, CF does not appear to be clever enough to know that we serve that same file whenever a file is not found (so that direct links to rooms work), and it doesn't seem to respect the max-age 60 that we set on the source bucket.

As a result, direct room links are cached by cloudfront for long periods of time, meaning they run an old version of the code. This should resolve that.


mrose17 (approved):
Thanks! I've wondered why sometimes I have to do a refresh after we deploy...

PR #161, opened 2022-01-25
Production Release: fully clear cloudfront on deploy 

tackley:
- #160


mrose17 (approved):
A-OK on staging, so promote to production

PR #181, opened 2022-02-16
Update jwt-store.ts

mrose17:


src/jwt-store.test.ts
```diff
@@ -98,7 +98,7 @@ test.only("jwt expiry works as expected", () => {
 
   expect(reloaded.findRefreshTokenForRoom("ok")).toBe(nonExpiredRefresh);
   // temporarily: the expired refresh token should not be removed
-  expect(reloaded.findRefreshTokenForRoom("expired")).toBe(expiredRefresh);
+  // expect(reloaded.findRefreshTokenForRoom("expired")).toBe(expiredRefresh);
```
tackley, line 101: I'm generally not a fan of committing commented out code, as it often serves to confuse the next person when they don't know _why_ it was commented out. (was it supposed to work, but didn't? is it temporarily disabled? is it something we're planning?)  We can obtain the old lines through git history if needed.

But in this case, there's a test we actually want to work: after gc, the expired token should be removed:

```typescript
expect(reloaded.findRefreshTokenForRoom("expired")).toBeUndefined();
```

mrose17: That's much better, thanks!

PR #219, opened 2022-03-23
Version bumps

tackley:
Including some minor npm audit fixes that would not have affected us.


mrose17 (approved):
Thank you dependabot!

PR #228, opened 2022-03-29
Production Release: ensure users who just use gcal extension don't lose subscription status

mrose17:
Working good on main!


tackley (approved):
Looks good as previously reviewed. And QA check passed on staging.

PR #245, opened 2022-04-08
Broken webpack link "https:://" => "https://"

NikolaiSch:
Link redirected to a local github file that didnt exist, changed to official webpack website


mrose17 (approved):
Good catch, many thanks!

PR #260, opened 2022-04-22
Rollup of dependabot updates as of 2022-04-22

mrose17:
Looks good on dev.


tackley (approved):
Since these changes are only to @types - used during tsc compilation only - and the tsc compilation passes, I'm happy with this without any additional checks.

Issue #98, opened 2021-12-15
Room is full message should link to the paid account sign up


jsecretan:
We should put an additional message and link, "Click here to sign up for a paid account and remove these limits" or some such and link that directly to the Brave Talk checkout.  This will help interested users know what to do next.

rmcfadden3:
We should avoid "click here" CTAs for accessibility reasons.

Suggested alternative:

- Header: Room is full
- Body / CTA: [Upgrade to Brave Talk Premium](https://account.brave.com/) to host calls with more people.

tackley:
How about we re-use the subscribe CTA panel from the home page?

![Screenshot 2021-12-16 at 10 05 48](https://user-images.githubusercontent.com/51444/146351358-34142bf4-8a51-41e5-8a9d-90222b68a7af.png)

stephendonner:
## Verified `PASSED` using `talk.brave.software/....`

### Steps:
1. started a call on `talk.brave.software/...`
2. joined from 4 clients
3. confirmed joining from the 5th yields "Room full" messaging with newly-added subscribe-CTA blurb

`talk.brave.software`

4 clients | 5th client, `room full`
---------|----------------------
<img width="1639" alt="Screen Shot 2022-06-02 at 10 35 50 AM" src="https://user-images.githubusercontent.com/387249/171694186-26bfb710-2749-4d09-ad83-9fe240991513.png"> | <img width="871" alt="Screen Shot 2022-06-02 at 10 36 08 AM" src="https://user-images.githubusercontent.com/387249/171694198-d2cc9a9f-7da2-4b7c-bc1f-49ef6c670184.png">


`talk.brave.com`

4 clients | 5th client, `room full`
---------|----------------------
<img width="1312" alt="Screen Shot 2022-06-02 at 10 38 16 AM" src="https://user-images.githubusercontent.com/387249/171693961-820ae6b9-365c-4ea6-a280-c7dc78e94f32.png"> | <img width="871" alt="Screen Shot 2022-06-02 at 10 38 28 AM" src="https://user-images.githubusercontent.com/387249/171693972-7ddfc084-87ea-4eac-8f5c-dfd851a35fac.png">

PR #312, opened 2022-05-31
Issue98

alanzhng:
cf., https://github.com/brave/brave-talk/issues/98

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -754,6 +754,9 @@ const joinConferenceRoom = async (
         ).href = `${subsUrl}/account/?intent=recover&product=talk`;
         subscribeCtaEl.style.display = "block";
       }
+
+      console.error(error);
+      notice(error.message);
     } else {
       console.error(error);
       notice(error.message);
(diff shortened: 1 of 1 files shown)


rmcfadden3:
@mrose17 — I think we should offer a little more copy so it doesn't feel like a complete dead end. Something like:

"Sorry, this room is full. Please contact the room creator."

Also: do we call these things "rooms" anywhere else on public-facing marketing / education materials? I don't recall seeing that word used. Instead, we've always called these units "calls." Will a user understand the term "room" in this case? If there's any concern they won't, we could say:

"Sorry, this call is full. Please contact the call creator."

(For reference, on https://brave.com/talk/ we say "calls" throughout, and in the first FAQ on that page we use the term "call creator" … so there's precedent for this latter suggestion.)

mrose17:
@alanzhng - Please use @rmcfadden3's text above:

        "Sorry, this call is full. Please contact the call creator."

and deploy to development for a final check...

@rmcfadden3 - thank you!

mrose17 (approved):
Looks fine on `dev` ... thanks!

PR #325, opened 2022-06-07
Issue11

hadiamjad:
Japanese language support for brave-talk:

This functionality works when you choose JP as your default browser language. So go to settings>language and make Japanese the default language. For all other languages, it will fall back to English. 

image


hadiamjad:
https://github.com/brave/security/issues/906

package.json
```diff
@@ -44,6 +44,7 @@
     "webpack-dev-server": "^4.9.0"
   },
   "dependencies": {
+    "i18next": "^21.8.6",
```
diracdeltas, line 47: we should be using exact versions instead of fuzzy versions in package.json (same for the other deps)

src/get-language-detector.ts
```diff
@@ -0,0 +1,7 @@
+export function getLangPref(): any {
+  var language =
```
diracdeltas, line 2: non-blocking: better to use `const`

src/get-language-detector.ts
```diff
@@ -0,0 +1,7 @@
...
+    (navigator.languages && navigator.languages[0]) || navigator.language;
+
+  if (language == "ja") return "ja";
+  return "en";
```
diracdeltas, line 6: non-blocking: could just write `return language === "ja" : language : "en"`

please use === instead of ==

diracdeltas:
i would suggest adding a pre-merge linter check to this PR to catch the kind of issues i pointed out above

diracdeltas:
> i would suggest adding a pre-merge linter check to this repo to catch the kind of issues i pointed out above

usually this is just done by adding an `npm lint` script which runs as a pre-merge github action.  most repos at Brave use https://www.npmjs.com/package/standard. this can be done as a separate PR since it will probably find a bunch of issues. cc @alanzhng

hadiamjad:
I've updated it. @diracdeltas

diracdeltas:
see comment above about pinning exact versions in package.json

hadiamjad:
> see comment above about pinning exact versions in package.json

updated! @diracdeltas

diracdeltas (approved):
this is fine but please open a follow up for https://github.com/brave/brave-talk/pull/325#issuecomment-1151442640, thanks

alanzhng:
@diracdeltas I've been looking into adding `standard` to this repo. I wanted to clarify if we should keep both `standard` and `Prettier` as code review / formatting tools.

diracdeltas:
@alanzhng definitely prefer `standard` because prettier doesn't catch code bugs. https://prettier.io/docs/en/comparison.html

diracdeltas:
in fact since `standard` also checks formatting, you can probably replace prettier entirely

Issue #314, opened 2022-06-01
Add additional context to a "Request failed: ..." message


alanzhng:
The current error screen lacks information on what went wrong. We should add context on why a request is failing and what the status code means.

![image](https://user-images.githubusercontent.com/40369305/171503249-408cd6b0-9e91-43dc-aaac-e34e3ca5d750.png)

Issue #11, opened 2021-10-01
Localize Brave Talk for JP


mrose17:
Step one:  find out the current status of adding i18n

hadiamjad:
cf https://bravesoftware.slack.com/archives/C01BC0K5QP7/p1633102605407400

PR #330, opened 2022-06-09
Added more failure messages (context to error status codes)

alanzhng:
c.f., https://github.com/brave/brave-talk/issues/314


alanzhng:
Added a couple new failureMessages that I felt were relevant to test. The messages are generic (e.g "Forbidden") and can definitely be changed to describe specifically what is going wrong with Talk / the subscription service.

@rmcfadden3, could you take a look and comment on the wording of these failure messages?

rmcfadden3:
@alanzhng — they seem ok, though note we should default to sentence case whenever possible. So edit to:

      403: "Forbidden",
      405: "Method not allowed",
      408: "Request timeout",
      417: "Expectation failed",
      420: "Method failure",
      429: "Too many requests",
      500: "Internal server error",
      502: "Bad gateway",
      503: "Service unavailable",
      504: "Gateway timeout",

Also, question: Are these Brave Talk errors? If yes, we had an earlier thread where we discussed defaulting to "Call" rather than "Room" for the atomic unit the user creates or participates in. You start a call, end a call, join a call, etc. So anyways if this is for Talk, would suggest changing the following as well:

      404: "The call does not exist",
      409: "Sorry, call already exists! (this should not happen)",

alanzhng:
Will do!

Yes, these are Brave Talk errors. I completely agree that "call" is the most clear and leaves zero room for confusion.

mrose17:
@alanzhng - great! the next step is to add the JP language localizations since the fix for https://github.com/brave/brave-talk/issues/11 is now in `main` ...

here's how to do that:

1. Click the `Update branch` button above and then get the latest version of `issue314`

mrose17:
@alanzhng - great! the next step is to add the JP language localizations since the fix for https://github.com/brave/brave-talk/issues/11 is now in `main` ...

here's how to do that:

1. Click the `Update branch` button above and then get the latest version of `issue314`
2. Take a look at https://github.com/brave/brave-talk/pull/333/files to see how @hadiamjad add support for the Status check link
3. Get the JP translations - abdul will explain the process

Good luck!

mrose17 (approved):
Good catch!

mrose17 (approved):
Looks good on the development server and to me.

Issue #335, opened 2022-06-11
Replace Prettier with StandardJS in this repo


alanzhng:
This repo uses `Prettier` as a code review tool. @diracdeltas brought to our attention, however, that a majority of other Brave repos use `Standard` as their code review tool.

`Standard` is a linter AND can function as a formatter while `Prettier` is only formatter. This means `Standard` fixes formatting just like `Prettier` but also catches code quality issues.
https://prettier.io/docs/en/comparison.html

Complete `Standard` documentation:
https://standardjs.com/

I believe we will have to use the TypeScript variant.
https://standardjs.com/#can-i-use-a-javascript-language-variant-like-flow-or-typescript

tackley:
Worth a try - let's run it against the codebase and check that its formatting rules don't make the code less readable.

Obviously this should replace prettier rather than live alongside it.

diracdeltas:
@alanzhng i think you can go ahead, pretty sure there is a one-line command you can run in Standard to fix those issues automatically

alanzhng:
Sounds good!

alanzhng:
We moved away from Standard as some of the `eslint` rules caused the codebase to be a bit too verbose. The automatic fixing and formatting of the codebase was also not ideal.

The new approach we took on is simply installing `eslint-typescript` and using its recommended configurations alongside a few others. `ESLint` works in tandem with `Prettier`, meaning that the `Prettier` formatting that some developers have become accustomed to remains. In general, `ESLint` is simply more configurable and seems to suit this repo's needs better than `Standard`.

mrose17:
@alanzhng - thanks. i think the key thing is that the `estlint-typescript` gives us the additional checking we want without making the code (much) harder to read.we've already done one big shuffle (when we moved to `prettier`) and using `standard` resulted in some very large diffs and every file being touched. i confess that after 45m of reviewing the diffs (and still not having reached the end), it became obvious that using `standard` was unfortunate.

PR #361, opened 2022-06-24
Add eslint-typescript to this repo

alanzhng:
c.f., https://github.com/brave/brave-talk/issues/335

Added `eslint-typescript` as a linter for this codebase. The linter works in tandem with `prettier` to maintain all formatting and style rules. 

Linting can be run by the command `npm run format`, and only reports errors/warnings (it does not automatically fix code). Linting is configured to ignore all `.js` files as well as ignore the `no-explicit-any` rule.

I disabled ESLint's `no-explicit-any` rule because of its relationship with Typescript's `no-implicit-any` rule; adhering to these rules would make the existing codebase unnecessarily complex in my opinion.

There's also the option of adding a pre-commit hook that lints the codebase, ensuring only linted code is committed and pushed to the remote repo. This is something that can be added easily if we choose to.


src/index.ts
```diff
@@ -615,8 +617,10 @@ const renderConferencePage = (roomName: string, jwt: string) => {
     }
   });
 
-  document.getElementById("talk")!.style.backgroundColor =
-    options.interfaceConfigOverwrite.DEFAULT_BACKGROUND;
+  let talkEleBgColor = document.getElementById("talk")?.style.backgroundColor;
```
mrose17, line 620: Q: `talkEleBgColor` is defined, but never used in this function. it seems to me that lines 620-623 could be removed. Do you agree?

.eslintignore
```diff
@@ -0,0 +1 @@
+*.js
```
mrose17, line 1: Which files are these? I think there's only one, `src/js/jwt-decode.js` ... i think it would be better to explicitly exclude that file ... that way, if other `.js` files are added in the future, warnings will show and a decision can be made as to whether those files should be excluded. wdyt?

src/index.ts
alanzhng: I did this to resolve the no-non-null-assertion error from ESLint.
https://palantir.github.io/tslint/rules/no-non-null-assertion/

I could always remove the `talkEleBgColor` variable and simply check if `document.getElementById("talk")?.style.backgroundColor` is non null if that makes it more clear!

.eslintignore
alanzhng: There are a few other config files in JS such as:
`jest.config.js`, `postcss.config.js`, and `webpack.config.js`

I do agree however that it would be more helpful to see the warnings of possible future `.js` files rather than ignoring all of them.

src/js/jwt-decode.js
```diff
@@ -1,3 +1,5 @@
+/* global define */
```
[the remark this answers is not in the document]
alanzhng: This comment tells ESLint that `define` is a global variable and thus solves the `no-undef` error.
https://eslint.org/docs/latest/rules/no-undef

I agree that it clutters the code, there is an option to turn off this error in `typeof` checks if necessary.

src/recordings-ui.tsx
```diff
@@ -1,7 +1,7 @@
 import "./css/recordings.css";
 import DownloadImage from "./images/download.svg";
 import MediaPlayerImage from "./images/media_player.svg";
-import React, { Dispatch } from "react";
+import React /* , { Dispatch } */ from "react";
```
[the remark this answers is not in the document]
alanzhng: `Dispatch` was declared but never used in this file. I figured there was some value in commenting it out rather than deleting it but I see how it clutters the code.

alanzhng:
In the process of solving some of the requested issues, I found and set the option of environments to encompass both `browser` and `node`, solving some of the issues that resulted from not ignoring all `.js` files.

I also added `webpack.config.js` to the ignore list because attempting to fix the `no-var-requires` errors lead to the project being unable to start. I figured it would be best to leave the file alone as it is an important configuration file.

src/index.ts
mrose17: right. as far as i can tell the variable is never used, so. removing the four lines would be best.

src/js/jwt-decode.js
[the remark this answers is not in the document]
mrose17: great, let's go for that!

mrose17:
Fixes https://github.com/brave/brave-talk/issues/335

src/recordings-ui.tsx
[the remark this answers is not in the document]
tackley: Just as an aside here: as @mrose17 hints at, the general rule is to *never* commit *any* commented out code. It usually just serves to confuse the next person as when they encounter it and try to figure out _why_ it was commented out. If it's not needed, delete it. We use source control to track changes, so we can always look in the source control history to find code that has been deleted if needed. 

Happy with your update to this, so no further changes needed: just though it might be useful to outline this general guideline for future reference ;)

.husky/pre-commit
```diff
@@ -2,3 +2,4 @@
 . "$(dirname "$0")/_/husky.sh"
 
 npx pretty-quick --staged
+npm run lint
```
tackley, line 5: It might be worth switching from `pretty-quick` to [`lint-staged`](https://github.com/okonet/lint-staged). This implementation here will run lint on the entire codebase on every commit, rather than just changed files.

However, since a full lint on this repo takes less than a second (on my laptop), this is probably fine for now and we can iterate on this later.

package.json
```diff
@@ -6,6 +6,7 @@
   "scripts": {
     "build": "NODE_ENV=production webpack --config webpack.config.js --mode production",
     "format": "prettier --write .",
+    "lint": "eslint .",
```
tackley, line 9:
```suggestion
    "lint": "eslint src/",
```

Because if you run `npm run build` or `npx tsc` both of these create temporary directories under the root (`html` and `tsbuild` respectively) which will then cause the linter to false fail. 

Either ensure all temporarily build directories are ignored; or (recommended) just lint the source code tree not other files.

src/recordings-store.ts
```diff
@@ -1,4 +1,4 @@
-import { loadLocalJwtStore } from "./jwt-store";
+// import { loadLocalJwtStore } from "./jwt-store";
```
tackley, line 1:
```suggestion

```

no commented out code please

tackley (changes requested):
Looking good, @alanzhng.

Three minor things that must be addressed before approval:
- `npm run build` followed by `npm run lint` should not fail (see comment on package.json)
- [this discussion](https://github.com/brave/brave-talk/pull/361#discussion_r907130706) needs resolution
- removal of the one last line of commented out code

src/index.ts
alanzhng: There were only two commits I could find using `git log` regarding these lines and both did not tell me much. 

Regarding the intent of the line, `talk` is the id of the `body` element in our index.html file. When `renderConferencePage` is called (by clicking the start call button for example), this line sets the background color of the `body` element to #3B3E4F. 

However, this `body` element is never actually visible in the home page or the call page. I observed that the background color of this element actually changed only by deleting all of the `body` element's children as well as the `head` element through the browser dev tools.

I think that these lines are safe to delete seeing as they serve to make only visual changes that aren't even visible.

tackley (approved):
Great work @alanzhng!

mrose17 (approved):
All changes approved. Looks good. Thank you!

PR #401, opened 2022-07-26
dismiss extension promo if it is crossed once

hadiamjad:


mrose17 (approved):
Tested on `dev` and looks good.

PR #405, opened 2022-07-27
Set `localSubject` in config

mrose17:
Should avoid the intermittent issue with the subject not being set to "Brave Talk".

Looks good on laptop and `dev`


hadiamjad (approved):
looks good

PR #519, opened 2022-10-04
Rightsize mobile resolution

mrose17:
Looks fine on laptop and `dev`


tackley (approved):
Looks good, as discussed on slack

PR #527, opened 2022-10-07
comment out userHasOptedInToAds...

mrose17:
...remove entirely once we are certain that "nothing breaks" (i'm certain now, but i'm erring on the side of ultra-caution)

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -43,9 +43,6 @@ i18next.init({
   },
 });
 
-const useBraveRequestAdsEnabledApi: boolean =
-  !!window.chrome && !!window.chrome.braveRequestAdsEnabled;
-
 const env = process.env.ENVIRONMENT ?? "local";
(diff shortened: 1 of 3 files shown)


tackley (approved):
Looks good.

I think we can also remove `startCallButtonPromptsOptIn` from `WelcomeScreenOptions` as that now appears to be always set to false. There's UI in the HTML and some supporting images that can also go as part of that. But very happy to reduce risk by implementing that in separate PR.

mrose17:
> Looks good.
> 
> I think we can also remove `startCallButtonPromptsOptIn` from `WelcomeScreenOptions` as that now appears to be always set to false. There's UI in the HTML and some supporting images that can also go as part of that. But very happy to reduce risk by implementing that in separate PR.

OK. merging this now and will then submit a separate PR.

PR #575, opened 2022-11-03
retry on a 409 when createP=true

mrose17:
I tested with hand-crafted URLs, but I suspect the only real test will be in production... _sigh_

### DIFF
--- a/src/rooms.ts
+++ b/src/rooms.ts
@@ -22,6 +22,7 @@ interface RoomResponse {
 interface RoomsRequestParams {
   roomName: string;
   urlSuffix?: string;
+  createP: boolean;
   method: RequestMethod;
   body: RoomRequestBody;
   successCodes: number[];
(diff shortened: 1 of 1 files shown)


src/rooms.ts
```diff
@@ -77,6 +79,19 @@ const roomsRequest = async ({
     console.log(`<<< ${method} ${url} ${status}`);
 
     if (!successCodes.includes(status)) {
+      if (createP && status === 409) {
+        console.log(`!!! retry ${method} ${url} ${status}`);
```
tackley, line 83: Do you not need to put a delay in here?  e.g. use a `setTimeout` to wait say half a second before retrying.  Just worried this might effectively "hang" by getting into a tight retry loop.

mrose17: i don't think it can loop because the retry has `createP=false` ... hence, if it fails a second time, it won't retry for a third...

tackley: Didn't spot that! Perfect :)

tackley (approved):

PR #643, opened 2022-12-20
Use backtick syntax when appropriate

mrose17:
Looks good on desktop and `dev`


tackley (approved):
Good change, that make the code a little easier to read and work with.

mrose17:
Thanks! It's a small thing, but consistency is good, even in small things i suspect.

PR #664, opened 2023-01-06
small thing to declare string type

mrose17:
declare each hash value in `window.location.hash` as a `string`. i'm not sure why typescript occasionally complains (the array is generated by a string `.split`), but there's no harm in making it explicit.


tackley (approved):
Agree typescript's type inference should pick this up automatically without the need for an explicit type declaration.

But no objection to making it more explicit.

PR #665, opened 2023-01-06
On joining, open breakout room list if any exist 

mrose17:


tackley (approved):
Looks good as approved on staging PR.

PR #666, opened 2023-01-06
Allow notify messages after joining

mrose17:
allow `notice` to generate notifications during a conference

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -876,7 +876,7 @@ const getAutoOpenRoom = (): string | undefined => {
   }
 };
 
-const notifyUID = "notifyUID";
+const notifyUID = "brave-talk-notice";
 
 const notice = (text: string) => {
   if (JitsiMeetJS) {
--- a/src/index.ts
+++ b/src/index.ts
@@ -876,7 +876,22 @@ const getAutoOpenRoom = (): string | undefined => {
   }
 };
 
+const notifyUID = "notifyUID";
+
 const notice = (text: string) => {
+  if (JitsiMeetJS) {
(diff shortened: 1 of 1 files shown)


src/index.ts
```diff
@@ -876,7 +876,22 @@ const getAutoOpenRoom = (): string | undefined => {
   }
 };
 
+const notifyUID = "notifyUID";
```
tackley, line 879: Should this not actually be a unique identifier?  Or is this just ensuring that we only show one message from brave talk at a time - in which case would it not make more sense for the constant value to be "brave-talk-notice" or similar?

tackley (commented):
The code looks fine. 

Can you give an example of when this would be used?

mrose17:
> The code looks fine.
> 
> Can you give an example of when this would be used?

At the present time, we write to the JavaScript consult when there are exceptional issues. Nine times out of 10, the issue is not really relevant to the user. So this policy is fine. However, every now and then we really would like to tell the user that something has gone sideways. For the project that I am working on, it is also very useful as a small, debugging tool. Of course, if that project goes into production, anything of a debugging nature can be removed.

src/index.ts
mrose17: > Should this not actually be a unique identifier? Or is this just ensuring that we only show one message from brave talk at a time - in which case would it not make more sense for the constant value to be "brave-talk-notice" or similar?

I like your suggestion. It just needs to be a string that will not be reused, and something that is meaningful, and unlikely to be used by the gypsy internals, is a better choice.

tackley (approved):
Looks good!

PR #747, opened 2023-03-14
delete the giphy proxy

mrose17:
until pre-production issues are resolved.


tackley (approved):
<insert witty giphy here>

PR #774, opened 2023-03-27
Force in-call display full screen

tackley:
Even if inline styles are not working.

See https://bravesoftware.slack.com/archives/C01BC0K5QP7/p1679915227867859: in some cases it appears the urls with room names in them are served with an old CSP that doesn't permit inline styles.

This is a temporary workaround so that the in-call UI still displays full screen even if inline styles are not permitted.


mrose17 (approved):
After review, emergency fix approved. Investigation is on-going as to why the cloud caching is behaving oddly.

tackley:
Not needed in the end as we've re-fixed the CSP.

PR #785, opened 2023-04-03
Use different link for external api JS

mrose17:
Dynamically set external api JS based on tenant ID

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -51,7 +51,7 @@
       content="https://talk.brave.com/images/brave-talk-tweet.jpg"
     />
 
-    <script src="https://8x8.vc/libs/external_api.min.js"></script>
+    <script src="https://8x8.vc/external_api.js"></script>
   </head>
 
   <body>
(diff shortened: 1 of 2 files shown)


tackley (commented):
This isn't going to work reliably, as at no point does the code await the promise returned by `miniLoadExternalApi`. That means when the code in `jitsi.ts` expects to find `JitsiMeetExternalAPI` in the global namespace, you have no guarantee that the script will have finished loading by that point. It will crash at that point.

mrose17:
@tackley - you're right, and more importantly...it might as well be synchronous since nothing is going to work until it loads. accordingly, i've changed the code to synchronous. so, it behaves like the current code does (synchronous) but doesn't hard-code the URL...

tackley (approved):
I do think we can optimise further here, as it would be sensible to no block the loading of the home page waiting for this script to load.

But as this mirrors current behaviour, am happy to treat that as a separate issue.

PR #806, opened 2023-04-12
Web3 on reactify

johnhalbert:


mrose17 (approved):
Looks good on desktop, and ready for `dev` and final integration and initial testing.

package.json
```diff
@@ -47,6 +47,9 @@
   },
   "dependencies": {
     "@emotion/react": "11.10.6",
+    "apg-js": "^4.1.3",
```
kdenhartog, line 50: Looks like the other dependencies are version locked for this repo - can we do the same for these additional dependencies?

src/components/web3/api.ts
```diff
@@ -0,0 +1,309 @@
...
+  const nonce = await getNonce();
+  console.log("!!! nonce", nonce);
+  const message: EIP4361Message = {
+    domain: "talk.brave.com",
+    address: web3Address,
+    statement:
+      "Please sign this message so Brave Talk knows that you own this address",
+    uri: "https://talk.brave.com",
+    version: "1",
+    chainId: 1,
```
kdenhartog, line 179: I'm checking with the EIP-4361 authors what the expected behavior is here since we're not interacting with a chain here. It seems kind of odd to declare a chainId when the request is never being utilized by a chain.

src/components/web3/api.ts
```diff
@@ -0,0 +1,309 @@
...
+  const result = {
+    method: "EIP-4361-json",
+    proof: {
+      signer: web3Address,
+      signature: signature,
+      payload: hexPayload,
+    },
+  };
+
+  window.localStorage.setItem(STORAGE_KEY, JSON.stringify(result));
```
kdenhartog, line 201: What's the intent of storing this locally? Once it's been submitted we won't be able to reuse it because of the nonce replay protection we added. I think we can eliminate the `web3RestoreAuth` function all together too if we're not going to be storing this in localStorage.

src/hooks/use-web3-call-state.ts
```diff
@@ -0,0 +1,127 @@
+import { useState } from "react";
+import { Web3Authentication, web3Prove } from "../components/web3/api";
+import { POAP, NFTcollection } from "../components/web3/core";
+import { generateRoomName } from "../lib";
```
kdenhartog, line 4: Is this function still available? I'm not seeing it in `src/lib.ts` which is where this appears to be pointing to at the moment.

kdenhartog (approved):
Looking at the comments I left, I'm not seeing any specific blockers here. There's some stuff I'd like to dynamically test once we get this deployed to dev (specifically I want to see what happens when proxying the Simplehash requests and modifying the address before it hits our backend), but we other than that this all looks good to me

bcaller (approved):
Looks fine

package.json
mrose17: good catch!

src/hooks/use-web3-call-state.ts
johnhalbert: It is, here's the relevant lines: https://github.com/brave/brave-talk/blob/web3-on-reactify/src/lib.ts#L26-L34

package.json
johnhalbert: Good catch. Done!

src/hooks/use-web3-call-state.ts
kdenhartog: Ahh thanks, I wasn't reading the source tree properly.

tackley (commented):
A few minor nits, but especially removing the commented out code please.

src/components/web3/api.ts
```diff
@@ -0,0 +1,287 @@
...
+        return collections;
+      },
+      {}
+    );
+
+    return Object.values(collections);
+  };
+
+  try {
+    const getNFTsByWalletURL = `${SIMPLEHASH_PROXY_ROOT_URL}/api/v0/nfts/owners?chains=ethereum&wallet_addresses=${address}`;
```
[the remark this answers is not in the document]
johnhalbert: I've gone through and url encoded any input from the client for all of the URLs we call in the web3 api.

PR #834, opened 2023-04-21
Use buttons to select call permission type

tackley:
Which looks less ugly than a native drop down, and involves one less click.

https://user-images.githubusercontent.com/51444/233692943-3f6dee60-a769-4964-9723-52a3fb5d2010.mov


mrose17 (approved):
Much cleaner, thank you!

PR #859, opened 2023-04-25
Use WakeLock API to keep screen on in calls

johnhalbert:
This change makes use of the [WakeLock API](https://developer.mozilla.org/en-US/docs/Web/API/WakeLock) to keep the screen  on during calls.  Brave desktop supports this for sure, and I believe our Android build supports it as well (currently untested, however).  I do not know about support for iOS.

The purpose of this change is to keep the screen on during calls.  The design is such that the WakeLock API is utilized during the call only.  A wakelock is taken when the user joins the conference room, and is released when the user leaves.  If the user is on the main Brave Talk page, or on the join page, the wakelock will not be active and the screen will still lock.  But while they're in the conference room, the wakelock will ensure that, as long as the tab is visible, the screen will not lock.  If the tab is _not_ visible, the lock is automatically released.  When the tab is made visible again, the lock is re-acquired.

I believe this is sorely needed, especially on mobile.  My experience with Android in specific is that the screen will lock, disabling the speaker and microphone.  This requires continually interacting with the device, even if you're a passive participant in the conference, which is less than ideal.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -9,6 +9,7 @@
       "version": "0.1.0",
       "dependencies": {
         "@emotion/react": "11.10.6",
+        "@types/dom-screen-wake-lock": "^1.0.1",
         "ethers": "6.3.0",
         "i18next": "22.4.15",
         "react": "18.2.0",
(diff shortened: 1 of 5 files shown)


mrose17 (approved):
This should work well on desktop and android. I like the fact that it is very straightforward to use.

PR #874, opened 2023-05-03
Removed unneccecary scroll bar in meeting room

johnmarco123:


mrose17 (dismissed):
A simple, straight-forward fix that is much appreciated!

In the future, it may be a little easier to include a "before & after" screen shot in the PR (in this case, the change is easy to see, but at least i can compare what you're seeing to what i'm seeing on one of our development servers).

Although approved, a merge is unlikely until next week as we are presently doing a lot of regression testing prior to a major update, and (even though this is a minor change), I would prefer to avoid any jitter in the process.

Thank you!

johnmarco123 (commented):
edited iframe positioning to fix unneccecary sidebar, removed the previous buggy fix as well.

johnmarco123 (commented):
added a trailing comma to the end of the new iframe selector

johnmarco123:
Before:
image
After:
image
I don't think anywhere else on brave talk has iframes so this change shouldn't effect any other pages...

mrose17:
Thanks! This looks like it's working. I'm going to run a few more tests...

srirambv:
Did a check as well. There's no scrollbar for `https://talk.brave.software` unless required.

mrose17 (approved):
This looks good to me as well. Approved. Merging soon!

mrose17:
One check did not run, despite multiple attempts. No reason given. As this PR consists of a CSS change to remove a scrollbar if not needed, and has been tested on desktop and `dev`. The merge is authorized.

PR #878, opened 2023-05-03
Add i18next support for web3 components

johnhalbert:


mrose17 (approved):
the code looks fine, but the JP localization file is incomplete (i.e., it is missing a number of keys). let's take this opportunity to have it fully fleshed out.

here's how to do that: i created a `utility` directory. go to that directory and run `./i18n-compare.js jp > file.json` that will create a complete localization file that we can send and they can look for occurrences of `"<<< ..."` and fix them.

this can occur independently of the code updates here, which LGTM. so ...

PR #893, opened 2023-05-10
Add web3 gating w/ query param

johnhalbert:
Disables `Web3CTA` component unless query param `web3` is present with a value (e.g. `https://talk.brave.com/?web3=true`).  This just enables us to release this to production and test in that environment prior to making the feature openly available.


mrose17 (approved):
Unremarkable.

PR #901, opened 2023-05-15
Revert "Add web3 gating w/ query param"

johnhalbert:
This reverts commit a5728ca7efd2da4d28a4f17b9c8a59ca78f1b492.


mrose17 (approved):
No longer needed as the web3 query is in production for pre-FCS testing, and for FCS, the query param must be removed.

Issue #945, opened 2023-05-30
figure out how to fix text selection on the home page


mrose17:
when the site was made "entirely react", the text seems to have become unselectable. very odd. text selection needs to work!

johnhalbert:
See #947

PR #949, opened 2023-05-31
Re-enable text selection

johnhalbert:
Re #945


hadiamjad (approved):
works for me

mrose17 (approved):
Works on `dev` and it is excellent!

PR #1002, opened 2023-06-16
Abdul sol nft gating

hadiamjad:


src/components/web3/api.ts
```diff
@@ -3,6 +3,7 @@ import { isProduction } from "../../environment";
 import { fetchWithTimeout } from "../../lib";
 import { NFTcollection, POAP, NFT } from "./core";
 import { EIP4361Message, createEIP4361Message } from "./EIP4361";
+import { Buffer } from "buffer";
```
mrose17, line 6: looking at https://github.com/brave/brave-talk/pull/1002#issuecomment-1595134687 the `socket-security` scanner doesn’t like `bufferutil` and `utf-8-validate` because they have native code. i am guessing that `bufferutil` is being dragged in due to this line. you import that because of  https://github.com/brave/brave-talk/blob/abdul-sol-nft-gating/src/components/web3/api.ts#L270 so i suggest you find another way of doing that and then you can get rid of line 6 above.

PR #1037, opened 2023-06-22
Override semver to version 7.5.2

johnhalbert:
- Addresses https://github.com/advisories/GHSA-c2qf-rxjj-qqgw
- Adds override for semver package to use 7.5.2 for all dependencies


mrose17 (approved):
Thank you John!

PR #1054, opened 2023-06-26
Use renovate for dependency updates

tackley:
1. Disable dependabot for normal dependency updates. Note the already-created dependabot PRs will need to be manually closed, and the "security" dependabot PRs will continue.

2. Use [renovate](https://docs.renovatebot.com/) instead.

The renovate configuration included here is based on that from ads-serve, and:

a. waits for updates to have been published for 4 days before raising a PR - this both reduces churn and also avoids the 3 day period where npm-published entries can be "unpublished"
b. combines all minor and patch updates into a single PR

There are many many options to fine tune what you want renovate to raise PRs for and how to group them, but the configuration docs are pretty impenetrable. Happy to help you fine tune to your needs.


mrose17 (approved):
Let's try renovate as it is proving useful in `ads-serve`

Issue #833, opened 2023-04-21
reduce impact of web3 dependencies on non-web3 users


tackley:
The dependencies needed for web3 calls increase the size of the js bundle from 285K to 650K (non-gzipped, so actual impact is less).

We should explore whether we can use [webpack's code splitting](https://webpack.js.org/guides/code-splitting/) to make this into a separate bundle that's only loaded when the user selects a web3 call. 

(The actual gzipped size may not warrant this complexity however.)

mrose17:
As Graham notes, it may not be necessary because of the gzipping. After FCS, we can see exactly how big this is.

bcaller:
Based on package.json in https://github.com/brave/brave-talk/compare/dev#diff-7ae45ad102eab3b6d7e7896acd08c427a9b25b346470d7bc6507b6481575d519 we add `apg-js`, `buffer`, `ethers`, `i18next`. All seem OK.

[apg-js](https://socket.dev/npm/package/apg-js): Not clear if this is actually used by the code, `dev/apg-lib.less` appears to get picked up by jsonToLess so can be found in the bundled JS.

[ethers](https://socket.dev/npm/package/ethers): Contains a lot more functionality than we need, but OK.

mrose17:
the unused libraries are removed! I haven't seen any issues. it is deployed on `dev`

hadiamjad:
I checked the file size of the main bundle.js and `ethers` adds ~28%. Here is the complete breakdown:
![image](https://github.com/brave/brave-talk/assets/46374292/7bc3ade0-6d7f-4b99-ab8f-2ad6d69d84f7)

However, the Gzip version of the bundled main is ~200kb.
`brave-talk % npx gzip-size-cli /Users/haadi/Downloads/brave-talk/html/static/main.2bcc0a095bea1eaaffc7.js.gz
200 kB`

Do we still want to do webpacks code splitting?

mrose17:
I think we should split it. However, let's hear from @tackley as to his thoughts before doing any "real work" (-;

tackley:
Thanks for the bundle analyser graphs @hadiamjad, that's exactly the data we needed.

If we can do it cleanly without too much additional complexity, we should do it I think, on the assumption that a significant proportion of users will not use the web3 functionality, at the moment at least. We might also be able to get some of our components/web3 code (31K non gzipped) in the web3 only bundle too.

We need to ensure that the web3 bundle only gets loaded when the user:
- joins a web3 call, OR
[...]

PR #1066, opened 2023-06-27
reduce impact of web3 dependencies on non-web3 users

hadiamjad:
I have tested it locally and it reduces the size of main.js 

re #833


tackley:
> Unfortunately, this doesn't benefit the user in the current state as _both_ bundles are loaded when I go to the home page:

Actually I may have misread the network log - I think the js file named "606.*" is just the subscriptions wasm wrapper, not the bundle you've split out. But if I proceed to trying to make a web3 call (by disabling the subscription check) I don't see the extra bundle load. Did you observe this or am I doing something wrong?

hadiamjad:
> > Unfortunately, this doesn't benefit the user in the current state as _both_ bundles are loaded when I go to the home page:
> 
> Actually I may have misread the network log - I think the js file named "606.*" is just the subscriptions wasm wrapper, not the bundle you've split out. But if I proceed to trying to make a web3 call (by disabling the subscription check) I don't see the extra bundle load. Did you observe this or am I doing something wrong?

You should see when you try to start the web3 call after choosing NFT, that's when we need library:
image

tackley (approved):
My mistake - it appears the ethers bundle does only load when starting a web3 call. Thanks for testing this @hadiamjad!

PR #1082, opened 2023-07-03
Duplicate Fix For Update jest monorepo to v29 (major) 

hadiamjad:
This [PR](https://github.com/brave/brave-talk/pull/1065) is failing because `jest-environment-jsdom` is no longer shipped by default in the latest version.
**Solution:** we have to separately install `npm install jest-environment-jsdom`


mrose17 (approved):
Thanks for researching this!

PR #1102, opened 2023-07-06
Adding a check-box to filter NFTs with high spam score

hadiamjad:
checked:
image
unchecked (default):
image


mrose17 (approved):
Looks good on `dev1

PR #1108, opened 2023-07-11
AvatarURL persistance issues fix for Web3 Brave Talk

davidsuh11:
This is the PR for fixing the persistance issues for avatars in Web3 Brave Talk, in particular:
* Remove localStorage persistance so that users using a normal call after a Web3 call do not have persisted NFT avatars.
* Adding support for avatarURL information in JWTs so that avatars for each room is individually persisted. 
There are corresponding additions to `subscriptions` for this feature (https://github.com/brave-intl/subscriptions/pull/395).


src/jitsi/conference-page.ts
```diff
@@ -17,6 +17,26 @@ import {
  * @param {JitsiOptions} options - Configuration options for the conference room
  * @returns {IJitsiMeetApi} The Jitsi API instance for the conference room
  */
+
+function clearAvatarInfoFromLocalStorage() {
+  const jitsiLocalStorageSettings =
```
mrose17, line 22: calls to `localStorage.getItem/setItem` should be under a `try/catch` block.

mrose17 (approved):
Looks good on `dev2`

PR #1125, opened 2023-07-14
Add development3 env

mschfh:


mrose17 (approved):
This is just a `s/dev2/dev3/g` & `s/development2/development3/g` of an existing file that works great, so... APPROVED!

PR #1138, opened 2023-07-14
Update App.tsx -- fix for fallback route option

hadiamjad:
A new dependency i.e., Route was added and did not have a fallback option. So initially it was designed just to look / or ?sol therefore when a user tries to join a call with /{room_name}, it was unable to see which route it should take but now it has default route * that automatically solves problem.
Fix for this issue: https://github.com/brave/brave-talk/issues/1131

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -106,7 +106,7 @@ export const App: React.FC = () => {
           />
         )}
         <Route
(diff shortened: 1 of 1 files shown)


mrose17 (approved):
This should do the trick.

PR #1144, opened 2023-07-17
Arithmetic test for spam_score should be >= 80 (not > 80)

mrose17:
Looks good on desktop. Unable to test on `dev3`...


davidsuh11 (approved):
LGTM!

PR #1150, opened 2023-07-19
Update environment.ts

mrose17:


tackley (approved):
nit: maybe just `env.startsWith("development")` might avoid the next PR like this?

But this is perfectly readable, so I'm fine with it.

PR #1151, opened 2023-07-19
Dsuh 943 bat gating

davidsuh11:
PR to add change for BAT gating that was missed before merged into `main`


mrose17 (approved):
Unused, removed, winner.

PR #1164, opened 2023-07-21
Avatar selection bug fix for Web3 Brave Talk

davidsuh11:
This provides a fix for an avatar selection bug that was introduced due to the spam filter checkbox, as well as a visual bug resulting from identifying NFTs by their image URL rather than their NFT id.


mrose17 (approved):
Looks good on `dev2`

PR #1167, opened 2023-07-24
BAT gating translations

davidsuh11:


mrose17 (approved):
That matches what I saw...

PR #1178, opened 2023-07-26
updated sol gating and rebased with main

hadiamjad:


.node-version
```diff
@@ -1 +1 @@
-18.17.0
+18.17.0
```
mrose17, line 1: Can we get the newline put back?

package.json
```diff
@@ -43,16 +43,17 @@
     "pretty-quick": "3.1.3",
     "style-loader": "3.3.3",
     "svgo-loader": "4.0.0",
+    "ts-jest": "29.1.1",
```
mrose17, line 46: can we re-order the dependencies so that only the new one `buffer` shows up in the diff?

public/branding-config-dev.json
```diff
@@ -1,10 +1,10 @@
 {
-  "backgroundColor": "#000000",
+  "X-backgroundColor": "#000000",
```
mrose17, line 2: can we get rid of the two changes in this file? I'm not sure how they keep showing up.

public/branding-config-dev.json
```diff
@@ -1,10 +1,10 @@
...
   "didPageUrl": "",
   "inviteDomain": "https://talk.brave.software",
   "logoClickUrl": "",
   "logoImageUrl": "https://talk.brave.software/images/brave_logo_dev.svg",
-  "premeetingBackground": "url(https://talk.brave.software/images/background.svg) no-repeat center fixed",
+  "X-premeetingBackground": "url(https://talk.brave.software/images/background.svg) no-repeat center fixed",
```
mrose17, line 7: ditto

mrose17 (approved):
LGTM - asking @tackley to take a quick look...

src/components/WelcomeScreen.tsx
```diff
@@ -74,13 +88,107 @@ export const WelcomeScreen: React.FC<Props> = ({
...
+
+      const logoStyle: React.CSSProperties = {
+        width: "50px",
+        height: "50px",
+        marginBottom: "10px",
+        borderRadius: "50%",
+      };
+
+      return (
+        <div style={popupContainerStyle}>
```
tackley, line 145: Please use the `css` prop provided by @emotion/react rather than the raw `style` prop. Why? Because emotion creates a css class and applies that rather than applying the styles directly to the element.

In this case it doesn't make much difference, other than for consistency, but in other cases especially for repeated elements, this can impact rendering time and page size significantly.

src/components/web3/PermissionTypeSelector.tsx
```diff
@@ -31,17 +32,26 @@ const styles = {
...
   const { t } = useTranslation();
   return (
     <div css={{ display: "flex" }}>
       <button
         onClick={() => setPermissionType("POAP")}
-        css={[styles.base, permissionType === "POAP" && styles.selected]}
+        css={[
+          styles.base,
+          permissionType === "POAP" && styles.selected,
+          web3Account === "SOL" ? disabledButton : null,
```
tackley, line 53: nit: two different ways of optionally including styles right next to each other is a bit jarring.

I have no real preference between `condition && style` or `condition ? style : null` but please use the same syntax especially when the conditions are right next to each other. Otherwise the next reader of this code has to look twice to understand why these are differently phrased.

src/components/web3/SolLogin.tsx
```diff
@@ -0,0 +1,80 @@
...
+  onAddressSelected,
+}) => {
+  const [notice, setNotice] = useState<JSX.Element | undefined>();
+  const { t } = useTranslation();
+
+  useEffect(() => {
+    try {
+      window.braveSolana?.on("!!! accountChanged", () => setNotice(undefined));
+    } catch {
+      console.warn("!!! Brave Wallet does not exists");
```
tackley, line 22: nit: 

```suggestion
      console.warn("!!! Brave Wallet does not exist");
```

src/components/web3/SolLogin.tsx
```diff
@@ -0,0 +1,80 @@
...
+      setNotice(<span>{t("wallet_connect_request")}</span>);
+      web3LoginSol()
+        .then((address) => {
+          onAddressSelected(address, "login");
+          setNotice(undefined);
+        })
+        .catch((err) => {
+          console.error("!!! web3 window.braveSolana.connect", err);
+          setNotice(
+            <Trans i18nKey="wallet_connect_failed">
```
tackley, line 45: I'm not seeing this key added to `translations.json` below - is this correct?

src/jitsi/options.ts
```diff
@@ -182,6 +182,7 @@ export const jitsiOptions = (
   };
 
   const features = jwt_decode(jwt)?.context?.features;
+  features.recording = "true";
```
tackley, line 185: Is this change intentional? It doesn't seem related to adding sol gating?

hadiamjad:
@tackley  I've fixed ur suggested changes

Issue #1123, opened 2023-07-13
Add a "hidden" page on Brave Talk to aid in reporting "SPAMmy" NFTs


mrose17:
An amusing, but not unexpected, thing: one of my addresses got its first Polygon NFT yesterday and within 24h, 12 "SPAMmy" NFTs were added. About half had `spam_score >= 80` so weren't shown by default.
![Screen Shot 2023-07-13 at 14 54 34](https://github.com/brave/brave-talk/assets/400552/c9e28584-8b3b-45f9-8586-8f123bfa4aad)

I want to be able to click on something on this tab that will open another tab (that will be useful only to internal users).

This new page will show all the NFTs with their `spam_score` underneath and a checkbox. There will also be a "Submit" button at the bottom. If the user clicks the "Submit" button, then for each NFT that is checked will have its `collection.collection_id` listed, e.g.,
```
c89a8a129eb7e6e3a6a0895ce9910455
...
```
That's it. The user will then copy paste this into a `devops` issue and they will run a script to report spam to simplehash.

mrose17:
Fixed by #1182

PR #1187, opened 2023-07-31
Adding a hidden debug page for spam NFTs

davidsuh11:
This PR addes a hidden option, which is accessable through adding query parameter `?debug=y` in the URL for brave talk. This allows users to toggle a checkmark entering debug mode, where they can view spam scores via mouse over, and also view all the NFTs' collection IDs and copy them to their clipboard for reporting purposes. Addressing #1123 .

### DIFF
--- a/src/components/web3/StartCall.tsx
+++ b/src/components/web3/StartCall.tsx
@@ -8,7 +8,9 @@ import { Login } from "./Login";
 import { OptionalSettings } from "./OptionalSettings";
 import { bodyText, header } from "./styles";
 import { useWeb3CallState } from "../../hooks/use-web3-call-state";
+import { useParams } from "../../hooks/use-params";
 import { TranslationKeys } from "../../i18n/i18next";
+import { NFTDebugPanel } from "./NFTDebugPanel";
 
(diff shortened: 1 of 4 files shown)


src/components/web3/StartCall.tsx
```diff
@@ -25,12 +27,15 @@ export const StartCall: React.FC<Props> = ({
   setJitsiContext,
   isSubscribed,
 }) => {
+  console.log(useParams());
```
tackley, line 30:
```suggestion
```

either remove this log or add some extra text so people reading the console understand what is being logged and why

tackley (commented):
Looks great, other than the rogue `console.log`

src/components/web3/StartCall.tsx
mrose17: @davidsuh11 - i agree with @tackley, we really don't need this particular call to `console.log` ... if you agree, please remove it and i'll approve and we'll get it into `main` ... thanks!

mrose17 (approved):
Looks good on `dev2`

PR #1198, opened 2023-08-07
Implement web3 address exclusion behind a feature flag

hadiamjad:
feature: https://github.com/brave/brave-talk/issues/920


package-lock.json
```diff
@@ -12997,7 +12997,7 @@
         "debug": "^4.1.0",
         "gensync": "^1.0.0-beta.2",
         "json5": "^2.2.2",
-        "semver": "7.5.4"
```
tackley, line 13000: I'm confused as to why package-lock.json has changed when there's no change to package.json?

Are these changes an intentional change (if so please explain why they're here in the PR description) or not (in which case please eliminate them from the PR)?

src/components/web3/ExceptionListPanel.tsx
```diff
@@ -0,0 +1,150 @@
...
+
+  const handleInputChange = (event: React.ChangeEvent<HTMLTextAreaElement>) => {
+    setInputText(event.target.value);
+  };
+
+  const validateAddresses = () => {
+    const delimitersRegex = /,|\s|\t|\n/; // Regex to match comma, space, tab, or newline
+    const addresses = inputText
+      .split(delimitersRegex)
+      .map((address) => address.trim());
```
tackley, line 45: My general rule is that any regex should have a unit test to validate.

Regex's pro and their con is that they are very compact. This means they are very difficult to validate by inspection, and therefore difficult to maintain going forward.

Suggest you add a separate function like:

```typescript
export function splitAddresses(addr: string): string[] {
    return inputText
      .split(/,|\s|\t|\n/)
      .map((address) => address.trim());
}
```

Which would then enable you to write a simple jest test that validates the behaviour. Meaning that we have confidence that this regex is correct, and more importantly have the ability to safely modify it in the future when requirements change.

src/components/web3/ExceptionListPanel.tsx
```diff
@@ -0,0 +1,150 @@
...
+        !isValidAddress(address, web3Account) ||
+        compareList.includes(address) ||
+        address === web3Address
+      ) {
+        flag = 1;
+        invalidAddressesList.push(address);
+      } else {
+        validAddressesList.push(address);
+      }
+    });
```
tackley, line 64: nit: I prefer `.filter` and/or `.some` to a side-effecting `forEach` loop.

src/components/web3/ExceptionListPanel.tsx
```diff
@@ -0,0 +1,150 @@
...
+    if (flag === 1) {
+      setIsExceptionAddressWrong?.(true);
```
tackley, line 66: why the `?.` here? does this code behave as expected if `setIsExceptionAddressWrong` is not supplied? I'd prefer to make it a mandatory parameter that having an unusual optional call like this.

src/hooks/use-params.ts
```diff
@@ -11,6 +11,8 @@ interface Params {
 
   // url has "sol=y" meaning we should show option to start either solana or ethereum call
   isSolana: boolean;
+  // url has "sol=y" meaning we should show option for allowing address
+  isAllow: boolean;
```
tackley, line 15:
```suggestion
  // url has "allow=y" meaning we should show option for allowing address
```

Though perhaps naming the variable better (e.g. to `isAllowAddress` would eliminate the need for the comment).  Then it would not match the query string, but in fact `allow` is a fairly meaningless name for the query string too. So perhaps that should be `allow_address_exclusion` and the variable be `isAllowAddressExclusion`?

src/components/web3/StartCall.tsx
```diff
@@ -170,11 +183,18 @@ export const StartCall: React.FC<Props> = ({
...
               nftCollections={nftCollections}
               nft={nft}
+              exceptionList={exceptionList}
+              setExceptionList={setExceptionList}
+              allowList={allowList}
+              setAllowList={setAllowList}
+              isExceptionAddressWrong={isExceptionAddressWrong}
+              setIsExceptionAddressWrong={setIsExceptionAddressWrong}
               setPermissionType={setPermissionType}
               setNft={setNft}
```
tackley, line 199: There's now a _lot_ of params here. This may be not for this PR but I think we need to review how we're managing state as this is at the point of becoming an unmanageable mess.

Especially since we repeat it twice.

mrose17 (approved):
I like it! Let's wait for @tackley to take a look.

package-lock.json
mrose17: @tackley - sorry, i missed that.

@hadiamjad - you always have to look at these things to ensure that accidental updates don't get included. in particular, you need to get `package.json` and `package-lock.json` from `main`

src/components/web3/ExceptionListPanel.tsx
hadiamjad: Automatically it gets tested in Join call scenario

package-lock.json
tackley: This still need to be resolved - we'll end up reverting to an old version of `semver` and perhaps others when this PR is applied. Unless intentional this should be avoided.

I'm a little confused as to how this PR has 91 commits in it, dating back a very long time. It looks like work started on from a branch other than `main`?  

To resolve, I'd suggest you ensure this branch only includes commits necessary to deliver this feature. Once you've got into this state, usually an [interactive rebase](https://hackernoon.com/beginners-guide-to-interactive-rebasing-346a3f9c3a6d) is the way to remove all the irrelevant commits: on this branch, run `git fetch` followed by `git rebase --interactive origin/main`, then in the editor window that appears delete every commit that was not part of this work. That removes them from this branch's history. 

Finally, after checking it all builds, runs, and passes tests locally as usual, do a `git push --force-with-lease` to update this branch.

src/components/web3/ExceptionListPanel.test.ts
```diff
@@ -0,0 +1,40 @@
+import { splitAddresses } from "./api";
```
tackley, line 1: if the code is in `./api`, the tests should be in `api.test.ts` not here. Just renaming this file is fine.

src/hooks/use-params.ts
```diff
@@ -23,6 +25,7 @@ export function useParams(): Params {
       isCreate: p.get("create") === "y",
       isCreateOnly: p.get("create_only") === "y",
       isSolana: p.get("sol") === "y",
+      isAllowAddress: p.get("isAllowAddressExclusion") === "y",
```
tackley, line 28: usual convention is to use snake_case for query parameters, not camelCase

hadiamjad:
I have fixed these issues in this new PR: https://github.com/brave/brave-talk/pull/1215. so we can ignore this.

PR #1221, opened 2023-08-21
fix: correct prod location for labels-english

tackley:
as noted by @kdenhartog, this file should be sourced from talk.brave.com


mrose17 (approved):
Ouch! Good catch @kdenhartog ... and thanks @tackley

PR #1226, opened 2023-08-29
Brave Talk Send Crypto

dvdhs:


src/components/web3/SendCryptoPopup.tsx
```diff
@@ -0,0 +1,476 @@
...
+interface CryptoPopupProps {
+  incomingRequests: CryptoTransactionParams[];
+  setIncomingRequests: React.Dispatch<
+    React.SetStateAction<CryptoTransactionParams[]>
+  >;
+  web3Address: string;
+  jitsi: IJitsiMeetApi;
+}
+
+export const CryptoRecievePopup: React.FC<CryptoPopupProps> = ({
```
johnhalbert, line 85: This is hard to follow.  I would decompose this into constituent components `CryptoRecievePopup`, `CryptoSendPopup`, `CryptoOOBPopup`, `CryptoWrapper`.

johnhalbert (changes requested):

src/components/web3/SendCryptoPopup.tsx
```diff
@@ -0,0 +1,663 @@
...
+  const [showing, setShowing] = useState(DEBUG_SHOW);
+  const [params, setParams] = useState({} as any);
+
+  const returnSIWEToOrigin = async () => {
+    let siwe: SIWEReturnParams;
+    try {
+      siwe = await generateSIWEForCrypto(
+        web3Address,
+        params,
+        `Please sign this message so that ${params.senderDisplayName} knows that you own this address, and may send you ${params.amount} ${params.token}.`
```
bcaller, line 213: We should apply simple validation to amount and token (not sure if here or earlier in message processing) and reject message if amount not an int(?) or token not in enum rather than displaying arbitrary text. Best place is probably somewhere before `setParams(incomingRequests[0]);` below.

bcaller (changes requested):

bcaller:
Non-security: Type numbers then press backspace and `NaN` will appear in the amount box. Then it is impossible to erase `NaN`. Also behaviour gets weird when typing giant numbers (probably due to javascript imprecision).

image

src/components/web3/SendCryptoPopup.tsx
```diff
@@ -0,0 +1,663 @@
...
+          <Divider />
+          <div css={popupContentCSS}>
+            <div css={{ fontSize: "16px" }}>
+              <span>Send: </span>
+              <input
+                type="text"
+                pattern="[0-9]*"
+                value={value}
+                css={inputCSS}
+                onChange={(e) => setValue(parseInt(e.target.value, 10))}
```
bcaller, line 370: Come to think of it, we mustn't use the javascript Number type:

```js
> parseInt("")
NaN
> parseInt("12345678901234567890")
12345678901234567000  // missing 890
> parseInt("123456789012345678901234567890")
1.2345678901234568e+29
> parseInt("1234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890")
Infinity
> Number.MAX_SAFE_INTEGER
9007199254740991
```

Perhaps better to use BigInt:

```js
> BigInt("")
0n
> BigInt("12345678901234567890")
12345678901234567890n
> BigInt("123456789012345678901234567890")
123456789012345678901234567890n
> BigInt("1234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890")
1234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890n
> BigInt("what?")
// Uncaught SyntaxError: Cannot convert what? to a BigInt
> BigInt("0b10110")
22n
> BigInt("0XFF").toString()
'255'
> JSON.stringify(BigInt("1"))
// Uncaught TypeError: Do not know how to serialize a BigInt
> JSON.stringify(BigInt("1").toString())
'"1"'
```

davidsuh11: As a general response: We can note that the "units" in the input is actual *whole* units, not wei. For obvious reasons -- we dont want people to have to type 1e18 to send 1 BAT. I have gone ahead and addressed this by changing the input to a float type, and also there is a maximum allowed input (`Number.MAX_SAFE_INTEGER`). It seems hacky to just cap the send amount, but considering that (according to google) the max supply of BAT is 1.5B, it should be alright. Ethereum (for the case of futureproofing) has infinite supply, but theres ~100M in circulation at the moment, so this shouldn't be an issue in the future if there are expansions to the sending mechanism.

src/components/web3/SendCryptoPopup.tsx
```diff
@@ -0,0 +1,663 @@
...
+      const nonceExists =
+        outstandingRequests.filter((r) => r.nonce === nonce).length > 0;
+      if (!ok || !nonceExists) {
+        setCurrentResolution(null);
+        console.log("!!! not ok, bad signature");
+        return;
+      }
+
+      const currentPendingParams = outstandingRequests.filter(
+        (r) => r.nonce === nonce
```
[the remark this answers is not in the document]
davidsuh11: I've added the check to this part in 47a5e1c5d5afd3cc795e86ce010e96ace27a2bf8. Not sure if we need to run it again for the second case (L419-420 above).

package.json
```diff
@@ -43,22 +43,24 @@
...
     "typescript": "5.1.6",
-    "webpack-cli": "5.1.4",
     "webpack": "5.88.2",
-    "ts-jest": "29.1.1",
+    "webpack-cli": "5.1.4",
     "webpack-dev-server": "4.15.1"
   },
   "dependencies": {
     "@emotion/react": "11.11.1",
+    "@spruceid/siwe-parser": "^2.0.2",
```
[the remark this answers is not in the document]
davidsuh11: So the previous code only contains ways to generate a SIWE message, but we also need a way to unpack the SIWE message (in particular, the `nonce` field), which the library does. I suppose one could develop a parser in house thats a bit smaller, but I don't think the module is that big.

src/jitsi/event-handlers.ts
```diff
@@ -258,3 +268,55 @@ export const videoConferenceJoinedHandler = {
...
 };
+
+export const sendCryptoButtonPressedHandler = {
+  name: "participantMenuButtonClick",
+  fn: (jitsi: IJitsiMeetApi, context: JitsiContext) => async (params: any) => {
+    if (params.key != "send-crypto") return;
+
+    // add to outstanding messages
+    if (!cryptoAction.sendOutstandingRequest)
+      return console.error("!!! addOutstandingRequest not defined");
```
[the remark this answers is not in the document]
davidsuh11: This code no longer exists due to refactoring

src/components/web3/SendCryptoPopup.tsx
```diff
@@ -0,0 +1,663 @@
...
+  const [showing, setShowing] = useState(DEBUG_SHOW);
+  const [value, setValue] = useState(0);
+  const [pendingName, setPendingName] = useState("");
+  const sendAndRegisterOutgoingRequest = async () => {
+    const amount = value;
+    const token = "BAT";
+    const nonce = await getNonce();
+    const recipientDisplayName = (await jitsi.getRoomsInfo()).rooms
+      .filter((r: any) => r.isMainRoom)[0]
+      .participants.filter((p: any) => p.id === pending)[0].displayName;
```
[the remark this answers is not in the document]
davidsuh11: Swapped out in 47a5e1c5d5afd3cc795e86ce010e96ace27a2bf8

src/components/web3/SendCryptoPopup.tsx
bcaller: Hmmm OK, you're right I wasn't considering the wei at all.
However, both `parseInt` and `parseFloat` produce the same return type of `Number` so suffer from the same imprecision. Number is just not very suitable for dealing with precise amounts of money or big numbers or small numbers or combinations of big and small numbers...

```js
> parseFloat("100.000000000000007")  // 100000000000000007000 wei
100  // 7000 wei lost
```

Not the end of the world 💸 , but we probably should be using a combination of `parseUnits` and `formatUnits` as described in https://docs.ethers.org/v5/api/utils/display-logic/ to

> convert between user-friendly strings (usually displaying ~ether~BAT) and the machine-readable values that contracts and maths depend on (usually in wei)

src/components/web3/SendCryptoPopup.tsx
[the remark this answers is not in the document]
bcaller: The check you made here is good (and sufficient).

src/components/web3/SendCryptoPopup.tsx
```diff
@@ -364,10 +379,19 @@ export const CryptoSendPopup: React.FC<CryptoSendPopupProps> = ({
               <span>Send: </span>
               <input
                 type="text"
-                pattern="[0-9]*"
-                value={value}
+                pattern="\d*.?\d*"
```
bcaller, line 382: `.` matches everything which will probably allow `NaN` through again, needs escaping with `\.`

src/components/web3/SendCryptoPopup.tsx
```diff
@@ -125,6 +128,18 @@ const highlightBoxCSS = css`
   padding: 16px;
 `;
 
+const onlyNumberOrDecimal = (e: React.KeyboardEvent<HTMLInputElement>) => {
+  const charCode = e.key.charCodeAt(0);
+  if (
+    charCode > 31 &&
+    (charCode < 48 || charCode > 57) &&
+    charCode !== 46 &&
+    e.key != "Backspace"
```
bcaller, line 137: I don't think we should forbid backspace in case the user types something incorrectly (I've not actually tested it so not certain what happens).
Does this still allow pasting from the clipboard?
Perhaps we just don't show NaN when the textbox is erased?

bcaller (changes requested):

src/components/web3/SendCryptoPopup.tsx
```diff
@@ -0,0 +1,663 @@
...
+
+  // TODO: wrap in try catch
+  const resolvePending = async () => {
+    if (!currentResolution) return console.log("!!! currentResolution not set");
+    const { proof } = currentResolution.siwe;
+    const { signer, signature, payload } = proof;
+    const payloadStr = Buffer.from(payload.slice(2), "hex").toString("utf8");
+    const msg = new SiweMessage(payloadStr);
+    const currentRequest = outstandingRequests.filter(
+      (r) => r.nonce === msg.nonce
```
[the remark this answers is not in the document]
bcaller: Addressed below, thanks

package.json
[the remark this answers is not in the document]
mrose17: i'd rather use an existing library than write a new parser...

src/components/web3/SendCryptoPopup.tsx
bcaller: OK I think I misread the backspace part, but still I'm not sure this is what we should be doing.

davidsuh11: The code essentially only allows digits + `.` to be typed in the input box. You are right that this prevents pasting, which I will look into. 651df37ff44c8c4e31cd9653c98e6e79bd4412d7 has additional checking on the value, in particular, the `value` is now a string and the popup will refuse to send an outgoing request if `value` is not a proper (parsable) string.

src/components/InCall.tsx
```diff
@@ -17,15 +17,22 @@ import {
   subjectChangeHandler,
   videoQualityChangeHandler,
   videoConferenceJoinedHandler,
+  sendCryptoButtonPressedHandler,
+  //onEndpointTextMessageForCryptoHandler,
+  // onEndpointTextMessageCryptoSendReturned,
```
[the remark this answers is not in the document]
davidsuh11: addressed in 651df37ff44c8c4e31cd9653c98e6e79bd4412d7

src/components/web3/SendCryptoPopup.tsx
davidsuh11: This change has been made in 651df37ff44c8c4e31cd9653c98e6e79bd4412d7. The four components are their own files, and also definitions/css/components that are common to them are in `common.ts`, `common-css.ts`, and `common-components.tsx` resp. These are all in a new folder `web3/crypto`.

src/components/web3/SendCryptoPopup.tsx
davidsuh11: 651df37ff44c8c4e31cd9653c98e6e79bd4412d7 changes so that the transaction amount is a `string` until the call to `sendCrypto`, at which point we use `parseUnits` from ethers. There is a check on the outgoing request creation process to make sure that `amount` is a valid string, though I guess adding more try-catch's down the line couldn't hurt.

src/components/web3/SendCryptoPopup.tsx
davidsuh11: Fixed in [c9dbafc](https://github.com/brave/brave-talk/pull/1226/commits/c9dbafc9b7164daa55720f7d0c0c5b9c47ca4c29)

mrose17:
OBE: the code base has signficantly changed, so when work resumes on "Brave Talk Send Crypto", we should start with the `main` branch... The issues raised by QA are here https://github.com/brave/brave-talk/issues?q=is%3Aissue+label%3Asend-crypto+is%3Aclosed

mrose17:
*PLEASE DO NOT DELETE THE BRANCH!*

PR #1243, opened 2023-09-13
chore: configure eslint explicitly

tackley:
We used to rely on `eslint-config-react-app` to configure some aspects of eslint. This is the full-service config used by "Create React App". However it brings in other dependencies such as (an old version of) eslint itself. It also is much less maintained than it used to be, as many projects (including talk!) have moved away from using create-react-app.

Use the react specific plugins directly.

As part of this, eslint-plugin-react doesn't really like the style of declaring react functional components using `React.FC`. This is a bit of an outdated style anyway, and it's usually clearer just to create functions with explicit types. Since this is also how most react documentation sites do it, I've removed all uses of the `React.FC` type.


mrose17 (approved):
That's an awful lot of work. Thank you very much!

PR #1253, opened 2023-09-29
renovate.json: add org-level base

thypon:


bcaller (approved):
Just adds base config [`local>brave/renovate-config`](https://github.com/brave/renovate-config/blob/main/default.json)

PR #1263, opened 2023-10-05
#1260 is probably failing due to missing the `npm audit fix` for postcss

mrose17:
This supercedes #1260 which is failing, presumably due to `npm audit` reporting that there's a moderate vulnerability in `postcss`

@tackley - i'm not sure why `npm install` downgraded `semver` (take a look at `package-lock.json` ... should i revert those changes?

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -21,7 +21,7 @@
       "devDependencies": {
         "@brave-intl/skus-sdk": "0.1.3",
         "@types/jest": "29.5.5",
-        "@types/react": "18.2.23",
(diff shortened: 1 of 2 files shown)


tackley (approved):
I think the changes to `semver` in package-lock look ok: we're not actually using a different version, it's just the dependency requirements from other packages on it have changed. It still resolved to the same version.

mrose17:
Thanks for looking. I'm always very concerned when I see those kind of diffs (particularly the kind that are "hidden" in `package-lock.json`) so i appreciate you opining on the matter!

PR #1291, opened 2023-11-14
Fixes #1290 which updates prettier, but doesn't run prettier...

mrose17:
Looks good on desktop, `dev1`, and `dev3`

Looking at #1290 ... you'll see it fails because prettier is updated, but it isn't run, and so subsequent tests fail...


DJAndries (approved):
lgtm

PR #1324, opened 2024-01-05
Upgrade download github action to fix deploy failure

DJAndries:


mrose17 (approved):
I’m not sure why the path no longer needs a “./“ prefix, but hopefully that’s the fix…

PR #1344, opened 2024-01-25
Fix for #1341

mrose17:
The update to prettier causes `tsconfig.json` to be rejected, so fix that and update `package*.json`

### DIFF
--- a/package.json
+++ b/package.json
@@ -47,7 +47,7 @@
     "mini-css-extract-plugin": "2.7.7",
     "node-forge": "1.3.1",
     "npm-run-all": "4.1.5",
-    "prettier": "3.2.2",
+    "prettier": "^3.2.4",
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -47,7 +47,7 @@
     "mini-css-extract-plugin": "2.7.7",
     "node-forge": "1.3.1",
     "npm-run-all": "4.1.5",
-    "prettier": "3.2.2",
+    "prettier": "^3.2.4",
```
tackley, line 50:
```suggestion
    "prettier": "3.2.4",
```

Our standard is to use exact versions in package.json - and in fact you'll just end up with another PR from renovate to "pin" this version if you merge this.

bsclifton (approved):
++

tackley (approved):

PR #1360, opened 2024-02-09
Brave Talk Marvin to staging

mrose17:
Brave folks: see https://docs.google.com/document/d/1Cb15YelULNfjM50eanx6asdn4F354ufxkTPvtpNSZDM for the specification and https://github.com/brave-intl/brave-talk-server/pull/50 for the changes to the Brave Talk server

Briefly: a premium moderator may turn on transcriptions (speech-to-text) in addition to A/V recording. As with A/V recordings, the moderator has access to a transcription file that "lives" for 24h (with links on the home page). Unlike A/V recordings, all participants receive incremental transcription "chunks", which are visible to "Brave Leo", thereby allowing the participants to (1) ask what happened before they joined and (2) ask Leo questions related to the call.

PR #1367, opened 2024-02-21
Deploy Brave Talk Marvin to staging

DJAndries:
Supersedes #1360 , with a cleaner tree


tackley (approved):

src/components/InCall.tsx
```diff
@@ -38,26 +47,40 @@ export const InCall = ({
 }: Props) => {
   const divRef = useRef(null);
   const [jitsiMeet, setJitsiMeet] = useState<IJitsiMeetApi>();
+  const [transcript, setTranscript] = useState<string>("");
+  const transcriptManager = useRef(new TranscriptManager(setTranscript));
```
tackley, line 51: nit: you never seem to change this `ref`, so this can use the simpler `useState` or just `useMemo`.

src/components/InCall.tsx
```diff
@@ -79,5 +102,17 @@ export const InCall = ({
...
+    opacity: "0",
+    pointerEvents: "none" as const,
+    position: "fixed" as const,
+    zIndex: -1,
+  };
+
+  return (
+    <>
+      <div ref={divRef} css={{ height: "100%" }} />{" "}
+      <main style={hidden}>{transcript}</main>{" "}
```
tackley, line 115: nit: use `css=` here rather than style, as that will cause emotion to generate a css class for this rather than just inline stye. It would also avoid the need for the `as const` above if you wrapped the declaration in a `css` constructor. (You could also achieve that by just by inlining `hidden`.)

Not critical in this case though as the style is only used once, so the generation of named style classes by emotion won't have any performance improvement.

src/recordings-store.ts
```diff
@@ -1,11 +1,13 @@
...
-  url: string;
+  url?: string;
+  transcriptUrl?: string;
   roomName: string;
   createdAt: number;
-  ttl: number;
   expiresAt: number;
 }
 
+export const RECORDING_TTL = 86400;
```
tackley, line 9: nit: prefer e.g. `24 * 60 * 60` so I don't have to get a calculator out to understand how long this value is. Also would prefer this constant to have the unit in it as millisecond vs second vs minute confusion has tripped me up many times. e.g. `RECORDING_TTL_SECS`.

But making this a constant is much better than where I put it embedded insider the code somewhere :)

tackley (approved):
Looks great, with minor nits, none of which are sufficient to block, but please address if you agree.

src/transcripts.ts
```diff
@@ -0,0 +1,278 @@
...
+        timeout: "sticky",
+      });
+    }
+  }
+
+  delta2elapsed = (delta: number) => {
+    // i could not find a small JS package to do this....
+    const oneMinute = 60;
+    const oneHour = 60 * oneMinute;
+    const oneDay = 24 * oneHour;
```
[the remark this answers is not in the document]
mrose17: I like this idea, i'm moving it to `src/lib.ts`

src/recordings-store.ts
mrose17: agreed: better to let JS do the math (once) and to indicate the units in the constant's name.

src/transcripts.ts
```diff
@@ -0,0 +1,278 @@
...
+    }
+
+    this.preTranscript = await fetchTranscript(transcriptDetails.url);
+    this.updateTranscript();
+
+    return transcriptDetails.url;
+  }
+
+  // TODO(djandries): find a better name for this method, and for 'didT'
+  doT(jitsi?: IJitsiMeetApi) {
```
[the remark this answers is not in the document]
mrose17: ok, let's use "american" english for `initialize()` and `initializedP` ...

mrose17 (approved):
After (literally) weeks of testing, I approve this PR!

Disclaimer: I did some of this work myself...

Issue #1392, opened 2024-04-04
Add a "Leo" button to the toolbar


mrose17:
Add a "Leo" button to the toolbar. It should appear only for premium moderators. The button has two states: "on" and "off", which reflects whether transcription is on/off. The button toggles the transcription status. Note that this should not observe/interfere with A/V recording.

When a user (other than the moderator) gets their first transcript chunk, if they are using the Brave browser, use the Jitsi notification UI to tell the user that they can use Leo to analyze the meeting.

@minkyeongshin to provide the icon for the Leo button (based on what the Jitsi API expects).
@rmcfadden3 to provide exact wording on notification.

@bcaller - note that to detect the current state of recording, the site should listen for the `recordingStatusChanged` event.

rmcfadden3:
@minkyeongshin — can you drop a screenshot of your design mock once it's ready? That way I'll know how much real estate I have to work with, in terms of character count.

mrose17:
sorry for the confusion:  at the present time, the design "ask" is only for an icon to go in the toolbar:
![Screenshot 2024-04-05 at 08 15 51](https://github.com/brave/brave-talk/assets/400552/eefaf4ef-338e-4979-a9c9-85a338d59e89)
initially, we're planning to use the Jitsi [showNotification](https://jitsi.github.io/handbook/docs/dev-guide/dev-guide-iframe-commands#shownotification) API call. so you can have a "reasonable" amount of text (say up to 300 characters).

bcaller:
Is this actually possible with the JAAS API? https://developer.8x8.com/jaas/docs/customize-ui-buttons links to  [list of button names](https://github.com/jitsi/jitsi-meet/blob/a76f9d548b2cfceaf335250867b88a41667107e0/config.js#L782-L826)

bcaller:
Well we can brand the default icons as described at https://developer.8x8.com/jaas/docs/jaas-prefs-advanced-branding#branding-data-payload - but are we able to create a new icon or do we have to be sneaky and misuse an existing icon?

saghul:
https://github.com/jitsi/jitsi-meet/blob/a76f9d548b2cfceaf335250867b88a41667107e0/config.js#L949

You can use a data URL or any URL to an icon.

mrose17:
Thanks. Please remind me: what handler triggers when a button is pressed (presumably indexed by `id`)... sorry for not knowing this stuff better by now...

saghul:
[...]

PR #1393, opened 2024-04-09
Leo button

bcaller:
#1392


bcaller:
Needs:
1. Actual icon
2. Way to turn on transcription
3. Manage some toggle state so we can change it from Enable Transcription to Disable Transcription.

mrose17 (approved):
The size of the diff is a bit deceiving as we had to add some menu button changes, which we've never done before.

Regardless, looks good on `dev2`.

PR #1403, opened 2024-04-11
add default transcript

### LINKED ISSUE #1402: add default transcript
at present, the `<main/>` element is empty. here is the text to put into the element. if/when the first transcription chunk arrives, this text should be removed (depending on how the text is inserted, this may happen "automagically").

        Transcripts are not enabled for this call. You can ask the host to enable transcripts by going to "More options" and selecting "Start recording".

Based on experience, the actual text may change.

bcaller:
Closes #1402

### DIFF
--- a/src/components/InCall.tsx
+++ b/src/components/InCall.tsx
@@ -38,6 +38,8 @@ interface Props {
   jitsiContext: JitsiContext;
 }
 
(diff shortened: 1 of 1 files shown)


mrose17 (approved):
Very straight-forward. Note that because the `faster-whisper` model used by the JAAS is english-only, this isn't a i18n issue.

Issue #1409, opened 2024-04-16
Add transcript view page


DJAndries:
Implement the following transcript page (see the design in Figma):

![Screenshot-2024-04-16_09-23](https://github.com/brave/brave-talk/assets/5621455/0e9238c3-d4db-4815-bd0d-75e79bca788e)

- [ ] Add React route for transcript page (i.e. https://talk.brave.com/transcripts/<transcript id>)
- [ ] Fetch & parse text transcript from server
- [ ] Ensure "transcript available" notification points to new React page
- [ ] Ensure download button points to transcript text on server; add `Content-Disposition: attachment` header to server endpoint to prompt the Save As dialog

PR #1428, opened 2024-05-08
Transcript screen

bcaller:
Closes https://github.com/brave/brave-talk/issues/1409


src/App.tsx
```diff
@@ -53,18 +54,42 @@ export const App = () => {
     window.close();
   }
 
+  const initialiseRouteTranscript = () => {
```
DJAndries, line 57: nit: i know we're from the commonwealth, but US english is probably preferred :laughing:

src/App.tsx
```diff
@@ -53,18 +54,42 @@ export const App = () => {
...
+    const match = window.location.pathname.match(
+      /^\/transcript-([a-z0-9]{50})$/,
+    );
+    if (match) {
+      return match[1];
+    }
+  };
+  const [routeTranscript, setRouteTranscript] = React.useState<
```
DJAndries, line 65: maybe `routeTranscriptId` instead?

DJAndries (changes requested):

src/App.tsx
```diff
@@ -53,18 +54,42 @@ export const App = () => {
     window.close();
   }
 
+  const initialiseRouteTranscript = () => {
+    const match = window.location.pathname.match(
+      /^\/transcript-([a-z0-9]{50})$/,
```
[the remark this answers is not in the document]
bcaller: Something weird happens when there is a forward slash in the route: Visiting https://talk.brave.com/abc/def 302 redirects to https://talk.brave.com/

We would need to fix that first if we want forward slashes

src/App.tsx
bcaller: Changed to `init`

src/App.tsx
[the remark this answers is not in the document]
DJAndries: ah ok. probably CloudFront related. perhaps it'd be better to address as a follow up

src/components/Transcript.tsx
```diff
@@ -0,0 +1,347 @@
...
+      while (index > -1) {
+        const range = new Range();
+        range.setStart(textNode, index);
+        range.setEnd(textNode, index + str.length);
+        ranges.push(range);
+
+        index = content.indexOf(str, index + 1);
+      }
+    }
+    console.log(ranges);
```
DJAndries, line 112: looks like this can be removed

src/components/Transcript.tsx
```diff
@@ -0,0 +1,347 @@
...
+    const highlight = new Highlight(...ranges);
+    CSS.highlights.set("search-results", highlight);
+  }, [searchTerm]);
+
+  return (
+    <div
+      css={css`
+        margin: 93px auto;
```
DJAndries, line 120: perhaps we should keep most styling in `styles` above, so the tree is easier to read

src/components/Transcript.tsx
```diff
@@ -0,0 +1,347 @@
...
+              font-family: var(--Family-Headings, Poppins);
+              font-size: var(--Size-H2, 28px);
+              font-style: normal;
+              font-weight: 600;
+              line-height: var(--Line-height-H2, 36px);
+              letter-spacing: var(--Letter-spacing-Headings, -0.5px);
+              text-wrap: nowrap;
+            `}
+          >
+            Meeting Transcript
```
DJAndries, line 166: we should use react-i18next for any copy. i.e. https://github.com/brave/brave-talk/blob/0621b711728796ddc921fe88205eab5237086909/src/components/JoinCallSection.tsx#L36-L38

src/components/Transcript.tsx
```diff
@@ -0,0 +1,347 @@
...
+            justify-content: center;
+            align-items: center;
+            max-width: fit-content;
+          `}
+          kind="outline"
+          onClick={() => {
+            const link = document.createElement("a");
+            link.href = transcript.blobUrl;
+            link.download = `${transcript.id}.txt`;
+            link.click();
```
DJAndries, line 183: any pitfalls with this approach? we could just add a `Content-Disposition` header in the transcript response

src/components/Transcript.tsx
```diff
@@ -0,0 +1,347 @@
...
+              src={DownloadImage}
+              height="16"
+              width="18"
+              alt="download"
+              css={css`
+                margin-bottom: -2px;
+              `}
+            />
+          </span>
+          Download
```
DJAndries, line 197: also i18n

src/components/Transcript.tsx
```diff
@@ -0,0 +1,347 @@
...
+        range.setEnd(textNode, index + str.length);
+        ranges.push(range);
+
+        index = content.indexOf(str, index + 1);
+      }
+    }
+    console.log(ranges);
+    const highlight = new Highlight(...ranges);
+    CSS.highlights.set("search-results", highlight);
+  }, [searchTerm]);
```
DJAndries, line 115: we may want to add `transcript` to the  list of dependencies

src/downloaded-transcript.ts
```diff
@@ -0,0 +1,65 @@
...
+  for (const line of transcript.split("\n")) {
+    const timeOffset = line.match(/^(\d+m\d\ds) /)?.[1];
+    if (!timeOffset) continue;
+    const remainder = line.substring(timeOffset.length + 1);
+    for (const s in ACTION_NAMES) {
+      if (remainder.endsWith(s)) {
+        const result: TranscriptionEvent = {
+          timeOffset,
+          participant: remainder.substring(0, remainder.length - s.length),
+          message: s,
```
DJAndries, line 47: i18n as well

src/components/Transcript.tsx
bcaller: Do I need to add something in the JP translations, or do I just fill out the EN?

src/components/Transcript.tsx
bcaller: I am not aware of any issues with this. I believe this is the standard way to force a download of a file from javascript. The alternative is we need to make a change on the server-side with a parameter / route which adds the Content-Disposition header.

src/components/Transcript.tsx
DJAndries: I think English is fine for now

src/components/Transcript.tsx
DJAndries: okay. your call! will resolve for now since this seems okay

src/components/Transcript.tsx
bcaller: How do you trick it into displaying Japanese text anyway? I can't find a way

bcaller: Oh the 日本語のlocalisation has been broken for I assume a while.

https://github.com/brave/brave-talk/blob/23497eb4c89a7194e3af499cf8b59d7bc9dd1a69/src/get-language-detector.ts#L5

when I'm pretty sure we need to also accept `=== "ja-JP"`... that is what devtools sets it to when you pretend to be in Tokyo.

My `navigator.language` is `en-GB` for instance and not simply `en`.

bcaller: Fixed

image

DJAndries (approved):

PR #1462, opened 2024-07-16
Do not clear order URL parameters

DJAndries:


mrose17 (approved):
It looks like this fixes the problem.

PR #1467, opened 2024-07-20
Do not use payments staging in dev1

DJAndries:


mrose17 (approved):
This appears to provide the desired functionality!

PR #1471, opened 2024-07-29
Repeat double crypto pop fix for solana

bcaller:
Closes https://github.com/brave/brave-talk/issues/1236


bcaller:
@mrose17 OK should be fixed for Solana also now on dev2

mrose17 (approved):
we are getting closer!

PR #1485, opened 2024-08-23
Improve privacy by disabling JAAS requests to non-JAAS sites...

mrose17:
... e.g., for avatar URLs, which can be used to leak IP addresses.


mrose17:
Deferred until iOS updated to the latest SDK.

PR #1554, opened 2024-12-09
chore(audit): update path-to-regexp

tackley:
Only used in development by the webpack dev server, so the redos https://github.com/advisories/GHSA-rhx6-c78j-4q9w is not impactful, but upgrading to avoid the blocking npm audit failure.


tackley:
Force merging as this is a dev-only update, and runs fine locally.

PR #1598, opened 2025-02-28
Remove web3 functionality

DJAndries:


src/jwt-store.ts
```diff
@@ -208,7 +208,11 @@ const expiredP = (roomName: string, jwt: string): boolean => {
 
   try {
     const payload = jwt_decode(jwt);
-    return payload.exp < now;
+    const isWeb3 =
+      payload.context &&
+      payload.context["x-brave-features"] &&
+      payload.context["x-brave-features"].web3 === "true";
+    return payload.exp < now || isWeb3;
```
DJAndries, line 215: if normal or refresh JWT is a web3 one, mark it as "expired" so it will be removed during confab garbage collection

tackley (approved):
Looks great to me from inspection, and thanks for cleanly removing all the old code.

Could we also remove the now-unused dependencies from package.json? They won't be bundled so this isn't necessarily blocking to this PR, but let's get them out of the way.

Knip reports:

```
Unused dependencies (3)
@types/dom-screen-wake-lock  package.json:64:6
buffer                       package.json:65:6
ethers                       package.json:66:6
Unused devDependencies (1)
node-forge  package.json:49:6
```

(Though suspect `@types/dom-screen-wake-lock` may well actually be used and that's just knip not being able to see the usage properly.)

DJAndries:
> Could we also remove the now-unused dependencies from package.json?

I went ahead and removed ethers, thanks

PR #1617, opened 2025-03-26
chore: renovate config updates

tackley:
To reduce noise, and also ensure we are aware of major version updates:

1. rely on org-global settings ([here](https://github.com/brave/renovate-config/blob/main/default.json)) for mininum release age, rather than overriding
2. group all non-major updates (not just for npm)
3. re-enable PRs for major version updates, but show these only in the dependency dashboard for now (https://github.com/brave/brave-talk/issues/1058) as there are quite a few of these

### DIFF
--- a/renovate.json
+++ b/renovate.json
@@ -2,21 +2,16 @@
   "extends": [
     "local>brave/renovate-config",
     "config:js-app",
-    ":semanticCommitsDisabled"
+    ":semanticCommitsDisabled",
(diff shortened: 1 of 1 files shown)


mrose17 (approved):
I am always in favor of making things as simple as usable, and no simpler. Accordingly, I approve this update.

I hope that the `dependencyDashboardApproval` will be a help in keeping clutter to a minimum.

PR #1635, opened 2025-05-09
Transcript initialization error handling

DJAndries:
Resolves https://github.com/brave-intl/brave-talk-server/issues/135


mrose17 (approved):
This is a good fix for the root issue: a client-side snafu not initializing the transcript properly. I use the term _snafu_ because this is the first and only time we have seen this *ever*.

PR #1760, opened 2026-05-15
build(pnpm): set minimum release age

tackley:
Block any installs of deps less than 7 days old, and disallow non-repository dependencies.

Also ignore the two additional svelte audit issues, which we cannot avoid until `@brave/leo` upgrades the version of svelte it uses.


mrose17 (approved):
this PR improves dependency security and helps to avoid spiking my anxiety whenever I hear about a supply chain attack. given the near-immediate detection of package toxicity from Socket – thank you @feross & co. – this PR puts a "well-needed gap" between the attack and the automated PR system. _"my name is @mrose17, and I approve this request."_

PR #1763, opened 2026-05-22
pnpm audit --fix rollup

mrose17:
Looks good on desktop, `dev1`, `dev2`, and `dev3`.


mrose17:
The missing lines from `pnpm-workspace.yaml` were restored. i'm not sure how they got misplaced as I was preparing this rollup. Thankfully, it was caught!

tackley (approved):

PR #1766, opened 2026-05-29
Optimize recovery UI for non-premium users

DJAndries:
Only show the recovery token box on the homepage if the user is premium or is a new user/has no confab JWTs. Show a collapsed copy for free users.


tackley (approved):
A great improvement!

PR #1770, opened 2026-06-11
Only show "load tokens" option during recovery if there are no stored tokens

DJAndries:


mrose17 (approved):
Looks good on `dev1` (not possible to test on desktop, as desktop isn't able to be premium).

PR #1775, opened 2026-06-15
Cherry-pick dependency updates into prod

DJAndries:


mrose17 (dismissed):
This moves everything from `main` into `prod` except the Recovery Token code, which is what we want. Thank you!

mrose17 (approved):
Appears fixed.

PR #1782, opened 2026-06-22
Delete existing non-refresh JWTs when recovering rooms

DJAndries:


mrose17 (approved):
This makes the JWT store identical to what was there when the recovery token was created. This should be the desired behavior.

PR #1816, opened 2026-09-02
Address audit errors

DJAndries:


DJAndries:
pushed to dev2

tackley (approved):
I tend to manually clear out the old `minimumReleaseAgeExclude` entries (though latest versions of pnpm I think do this automatically), and tidy up the `overrides` list so it only contains one entry per component, just to make it easier to read.

But this is strictly optional and can be done subsequently, so I'm fine with this PR. It has the desired effect.

PR #1819, opened 2026-09-03
Update Brave talk 

aguscruiz:
## Summary

Restyles the Brave Talk homepage to the current Figma design and moves the whole welcome screen onto [Leo](https://github.com/brave/leo) design tokens. The page is now a stack of cards (join call, premium CTA, recovery token, recordings) instead of the old single hero, and every hard-coded colour, spacing, radius and font value in the touched components has been replaced with a `--leo-*` token.

Bundled Poppins and Inter font files are gone: typography now comes from Leo's `--leo-font-*` tokens, which removes 26 files and ~4.6 MB of `.ttf`/CSS from the repo.

Design reference: [📹 Brave Talk in Figma](https://www.figma.com/design/waBZo5gIAmaAYAygl0mssa/%F0%9F%93%B9-Brave-Talk) (homepage, recorded calls and transcript frames).

No functional or API changes — call setup, subscription checks, recovery-token handling and recording storage all behave exactly as before. This is presentation plus the copy changes needed for the new layout.

## What changed

**Typography and global styles**
- Deleted `src/fonts/*`, `src/css/poppins.css`, `src/css/inter.css` and their imports in `src/App.tsx`.
- `GlobalStyles` now sets the page background, base font and text colour from Leo tokens (`--leo-color-black`, `--leo-font-small-regular`, `--leo-color-primitive-neutral-95`).

**Layout and cards**
- `Section` is now the shared card shell: Leo radius, 1px neutral border, translucent neutral background, and responsive padding that tightens at ≤600px.
- `Header` uses the Brave icon + wordmark lockup, with the right-hand link rendering as a pill that switches between **Log in** and **My Account** based on subscription status.
- `JoinCallSection` was rebuilt: Talk logo, `h1` + subhead, a jumbo primary button, and the **Leo AI for meetings** promo block with the speech-bubble tail. On free plans it shows the "Up to 4 people. Unlimited with Premium." line; premium users get **Create link** instead. On ≤600px the layout stacks and centres, and the Leo promo block is hidden to keep the fold usable.
- `SubscriptionCTA` is now the **Premium calls** card with the trial button, and shows a spinner row while subscription status is still `unknown`.
- `RecoveryTokenBox` / `RecoveryTokenDialog`, `Recordings` and `Transcript` were restyled to match: pill-shaped outline actions with an inset border, an "Expires soon" badge on recordings, full-width actions on mobile, and a search field on the transcript view.

**Buttons and links**
- `Button` gained `variant` (`default` / `light` / `hero` / `outline` / `plain`) and `size` (`default` / `large` / `jumbo`) props, all token-driven, replacing the previous `hollow` boolean-only styling. Disabled state now uses `cursor: wait` with reduced opacity.
- Link styling is centralised on `Section` rather than repeated inline on each anchor, so all in-card links share one colour and underline treatment.
- Localised links use `<Trans>` instead of concatenated strings, so `recovery_token_learn_more` is now `<0>Learn more</0>.` — this keeps the trailing period outside the link in every locale.

**Progress and accessibility**
- While a call is being established, progress messages take over the button label inside an `aria-live="polite"` span; anything left over afterwards (e.g. an error) falls back to the subhead. Previously the notice and button label could show conflicting states.

**Assets and branding**
- Added `src/images/homepage/*` (Talk logo, Brave icon, wordmark, Leo icon and tail, background) plus `link-normal.svg`; refreshed `talkLogo.svg`.
- Removed 12 unused SVGs left over from the old homepage, wallet and copy-link UI (`solana`, `ethereum`, `plus`, `minus`, `close`, `content_copy`, `shape`, `media_player`, `papyrus`, `search`, `link_icon`, `download_color`, and the old dark logos).
- Refreshed the favicon, PWA icons (32/64/96/180/192) and the tweet preview image.
- Updated the Jitsi branding configs (`public/branding-config*.json`) to the current avatar palette and logo/background assets, so the in-call view matches the homepage.
- `webpack.config.cjs` now also copies Leo's `download` and `search` icons, which the recordings and transcript views use.

**Copy / i18n**
- New keys in `en` and `jp`: `free_call_limit`, `free_call_premium_link`, `leo_meetings_title`, `leo_meetings_description`, `premium_calls_description`, `start_trial_seven_days`, `transcript_search_placeholder`.
- Casing fixes to match the design: "Start premium call" (was "Start Premium call"), "Meeting transcript" (was "Meeting Transcript").

## Screenshots

**Desktop — free plan**

Homepage on desktop, free plan

**Desktop — premium**

Premium hides the subscribe CTA, switches the header to **My Account**, and swaps in **Start premium call** plus **Create link**.

Homepage on desktop, premium subscriber

**Mobile — 390×844**

Cards stack, buttons go full width, and the Leo promo block is hidden.

Homepage at mobile width

## Test plan

Automated, all run locally against this branch:

- [x] `pnpm run check:build` (`tsc`) — no errors
- [x] `pnpm run check:lint` (`eslint src/`) — clean
- [x] `pnpm run check:format` (`prettier --check .`) — clean
- [x] `pnpm test` (`jest`) — 3 suites, 9 tests passing
- [x] `pnpm run build` — compiles; only the two pre-existing asset/entrypoint size warnings

Manual, on `pnpm start`:

- [x] Free plan: header shows **Log in**; join card shows **Start free call** with the "Up to 4 people. Unlimited with Premium." line; **Premium calls** card and recovery-token card render.
- [x] Premium: header shows **My Account**; join card shows **Start premium call** and **Create link**; subscribe CTA is hidden; recovery card shows the premium copy.
- [x] Mobile (390px): cards stack and centre, buttons are full width, Leo promo block hidden, no horizontal scroll.
- [x] Links (Premium, Learn more, Log in, privacy policy, Service status) share one colour/underline treatment and point at the right destinations.
- [ ] Reviewer check: start a real call and confirm the in-call Jitsi branding (logo, pre-meeting background, avatar colours) looks right with the updated `branding-config*.json`.
- [ ] Reviewer check: with at least one recording in local storage, confirm the recordings list, the "Expires soon" badge, download action and the transcript view (including search) look right on desktop and mobile.

The premium screenshot above was taken by temporarily forcing `useSubscribedStatus()` to `"yes"` locally; that change is not part of this branch.

## CI note

The `check` job fails on `check:audit` (`pnpm audit`), not on anything in this diff — `check:build`, `check:lint`, `check:format`, `pnpm test` and `pnpm run build` all pass. The reported advisories are against `main`'s dependency tree (`postcss-selector-parser`, `qs`, `fast-uri`, `browserslist`, `@humanfs/node`), i.e. exactly the packages covered by the `pnpm-workspace.yaml` overrides in #1816 and #1814. Once either of those lands and this branch is rebased, `check` should go green. Happy to pull those overrides in here instead if maintainers prefer.

## Notes and follow-ups

- The dotted-grid background (`src/images/homepage/background.svg`) is ~1.12 MB after SVGO optimisation. It remains the largest new asset and could be replaced with a CSS pattern in a follow-up.
- The recordings and transcript views are restyled but were not re-architected; they still read from the same local store.
- Dependency and audit changes that were sitting on `dev` alongside this work (`webpack-dev-server` 6, `pnpm-workspace.yaml` overrides, lockfile) are deliberately excluded from this PR so the diff is presentation-only. They can land separately.


src/images/homepage/leo-tail.svg
```diff
@@ -0,0 +1,3 @@
+<svg preserveAspectRatio="none" overflow="visible" style="display: block;" width="32" height="32" viewBox="0 0 32 32" fill="none" xmlns="http://www.w3.org/2000/svg">
+<path id="Ellipse 3" d="M0 32C17.6731 32 32 17.6731 32 0V32H0Z" fill="#252527"/>
```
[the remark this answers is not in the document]
aguscruiz: SVGs are already run through svgo-loader at build time, so optimizing the source files should be behavior-neutral.

src/images/homepage/talk-logo.svg
```diff
@@ -0,0 +1,14 @@
+<svg preserveAspectRatio="none" overflow="visible" style="display: block;" width="72" height="72" viewBox="0 0 72 72" fill="none" xmlns="http://www.w3.org/2000/svg">
+<g id="origin-release">
+<path id="Bg" d="M0 28.0976C0 18.2625 0 13.345 1.91403 9.58846C3.59766 6.28415 6.28415 3.59766 9.58846 1.91403C13.345 0 18.2625 0 28.0976 0H43.9024C53.7375 0 58.655 0 62.4115 1.91403C65.7158 3.59766 68.4023 6.28415 70.086 9.58846C72 13.345 72 18.2625 72 28.0976V43.9024C72 53.7375 72 58.655 70.086 62.4115C68.4023 65.7158 65.7158 68.4023 62.4115 70.086C58.655 72 53.7375 72 43.9024 72H28.0976C18.2625 72 13.345 72 9.58846 70.086C6.28415 68.4023 3.59766 65.7158 1.91403 62.4115C0 58.655 0 53.7375 0 43.9024V28.0976Z" fill="url(#paint0_linear_0_520)"/>
+<g id="Product icon">
```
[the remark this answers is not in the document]
aguscruiz: SVGs are already run through svgo-loader at build time, so optimizing the source files should be behavior-neutral.

src/images/homepage/leo-icon.svg
```diff
@@ -0,0 +1,5 @@
+<svg preserveAspectRatio="none" overflow="visible" style="display: block;" width="32" height="32" viewBox="0 0 32 32" fill="none" xmlns="http://www.w3.org/2000/svg">
+<g id="Product icon">
```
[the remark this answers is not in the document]
aguscruiz: SVGs are already run through svgo-loader at build time, so optimizing the source files should be behavior-neutral.

src/images/homepage/brave-wordmark.svg
```diff
@@ -0,0 +1,9 @@
+<svg preserveAspectRatio="none" overflow="visible" style="display: block;" width="85.6802" height="24" viewBox="0 0 85.6802 24" fill="none" xmlns="http://www.w3.org/2000/svg">
+<g id="Brand / Brave Wordmark">
```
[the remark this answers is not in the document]
aguscruiz: SVGs are already run through svgo-loader at build time, so optimizing the source files should be behavior-neutral.

src/images/homepage/brave-icon.svg
```diff
@@ -0,0 +1,16 @@
+<svg preserveAspectRatio="none" overflow="visible" style="display: block;" width="35" height="40" viewBox="0 0 35 40" fill="none" xmlns="http://www.w3.org/2000/svg">
+<g id="Brand / Brave Icon / Release">
```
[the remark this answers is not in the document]
aguscruiz: SVGs are already run through svgo-loader at build time, so optimizing the source files should be behavior-neutral.

aguscruiz:
Pushed as `12713f6` to `restyling`, so it's now on [PR #1819](https://github.com/brave/brave-talk/pull/1819). All checks pass locally: `tsc`, `eslint`, `prettier --check`, and 9 tests across 3 suites.

Here are the seven fixes, all from real findings in the review comment:

**Localization.** The recordings card had five English strings baked in — the "Your recorded calls" heading, its description, the "Expires soon" badge, and the Transcript/Download action labels. Those now go through `t()` with new `recordings_*` keys in both `en` and `jp`. The Download action reuses the existing `download_transcript_button` key rather than adding a duplicate. Japanese was also missing `transcript_search_placeholder` and `Meeting Transcript` entirely, so the transcript view was falling back to English for a locale we actually ship; both are added.

**The reused translation key.** The Leo promo blurb was rendering its "Learn more" link through `i18nKey="recovery_token_learn_more"`. The text happens to match today, but it would silently break the promo the moment that string gets reworded for the recovery flow. There's now a generic `learn_more` key, and `recovery_token_learn_more` stays where it belongs.

**Header flicker.** The account link rendered "Log in" while `subscribed` was still `"unknown"`, then swapped to "My Account" once the check resolved — so every premium user saw the wrong label flash on load. The link is now hidden until the status resolves, matching how `SubscriptionCTA` already handles that state with its spinner.

**Transcript click fall-through.** `getTranscriptDisplayPath` can throw, and the `catch` returned *before* `e.preventDefault()` ran. The comment claimed it prevented a crash, but the click would just fall through to the raw `href` and navigate away to the transcript JSON. `preventDefault()` and the propagation calls now run first, so a bad URL is a no-op as intended.

**Branding config fetch.** `fetch("/branding-config.json")` piped straight into `.json()` with no `response.ok` check, so an HTML error page would reach the JSON parser and reject with a confusing error. It now throws on a non-OK response, which the existing `.catch` handles by keeping the fallback palette.

**Duplicate avatar color.** `#066CFF` appeared twice in `avatarBackgrounds` in all three branding configs, leaving 13 distinct colors instead of 14. The second occurrence is now `#1B8EF5`. This matters a bit more than before, since the transcript speaker colors read from this list.

**Recordings timer.** The 60-second `setInterval` was created even when there were no recordings to render. It's now gated on having at least one; it already cleared on unmount.

I did not change eight other items in that comment, because I checked each and they aren't bugs. The most important one: the reviewer flagged the deleted fonts as having "no visible replacement," but `src/index.tsx` already imports `@brave/leo/tokens/css/variables.css`, and Brave ships Inter natively — re-bundling the TTFs would undo the 4.6 MB the PR removes. The Node version claim is also wrong: `.node-version` is already 24.19.0 and CI reads that file. The rest (keys only in en/jp, `talk_title_premium` unused, `SectionWithLogo` "orphaned", `cursor: wait`, the dialog's `position: absolute`, the prod config filename) I covered in more detail in my previous message.

src/components/Button.tsx
```diff
@@ -11,53 +13,149 @@ interface Props {
...
   base: css({
-    margin: "0 auto",
-    width: "377px",
-    borderRadius: "48px",
-    padding: "17px 30px 19px",
+    margin: "var(--leo-spacing-none) auto",
+    width: "min(377px, 100%)",
+    borderRadius: "var(--leo-radius-full)",
+    padding:
+      "calc(var(--leo-spacing-xl) + var(--leo-spacing-xs) / 2) calc(var(--leo-spacing-3xl) - var(--leo-spacing-xs)) calc(var(--leo-spacing-xl) + var(--leo-spacing-s) - var(--leo-spacing-xs) / 2)",
```
DJAndries: nit: this is a fairly complex property. not really a big deal, but would it be better to supply absolute values so that it's a little easier to tweak later on?

src/jitsi/options.ts
```diff
@@ -75,6 +75,9 @@ export const jitsiOptions = (
       hideEmailInSettings: true,
       inviteAppName: "Brave Talk",
       localSubject: "Brave Talk",
+      prejoinConfig: {
+        showHangUp: false,
```
DJAndries: just curious: why no hang up button? is it considered an eye sore?

src/i18n/locales/en/translation.json
```diff
@@ -59,10 +64,17 @@
...
+  "premium_calls_description": "Host video calls with hundreds of people, get moderator controls, secure your meetings with Leo AI assistant tailored to your private sessions without compromising confidentiality. No installation needed.",
+  "start_trial_seven_days": "Start 7 day free trial",
   "recovery_token_title": "Meeting recovery token",
   "recovery_token_description": "If you're the moderator of a recurring meeting, you may export your call's unique token each time you set up a new call, and keep this token in a safe place. You'll need this token to retain moderator privileges in cases where you need to reinstall the browser or install on a new device.",
   "recovery_token_description_free": "If you're the moderator of a recurring meeting and have a recovery token, you may recover rooms here.",
-  "recovery_token_learn_more": "Learn more.",
+  "learn_more": "<0>Learn more</0>.",
+  "recovery_token_learn_more": "<0>Learn more</0>.",
+  "recordings_title": "Your recorded calls",
+  "recordings_description": "Recorded calls are automatically cleared 24 hours after their recording time.",
```
DJAndries: nit: "Recorded calls are automatically deleted 24 hours after the time of recording."

src/i18n/locales/en/translation.json
```diff
@@ -19,11 +19,15 @@
...
   "footer_pst_text": "privacy policy",
   "footer_status_text": "Service status",
+  "free_call_limit": "Up to 4 people. Unlimited with",
+  "free_call_premium_link": "Premium",
   "identifier_fetch_error": "Failed to fetch member identifiers (NFTs/POAPs)",
   "invalid_token_error": "Access failure: You must have an approved token to join this call",
   "address_exclusion_error": "Access failure: Your address is not in the allow list to join this call",
   "join_call": "Joining...",
   "join_call_error": "Error joining room",
+  "leo_meetings_description": "Never take meeting notes again. Leo brings AI to your meetings with AI summaries. Get transcripts and use AI chat to get answers from your meetings, and more.",
```
DJAndries: should we replace "AI chat" with "Leo"?

DJAndries: maybe "Get transcripts" should be "Generate transcripts" or something similar

src/i18n/locales/en/translation.json
```diff
@@ -105,7 +117,8 @@
   "Sorry, the call is already full": "Sorry, the call is already full",
   "Sorry, this call is full. Please contact the call creator.": "Sorry, this call is full. Please contact the call creator.",
   "Sorry, you are not a subscriber": "Sorry, you are not a subscriber",
-  "Start Premium call": "Start Premium call",
+  "Start premium call": "Start premium call",
```
DJAndries: did you intend to make premium lowercase?

src/i18n/locales/en/translation.json
```diff
@@ -19,11 +19,15 @@
   "footer_pre_text": "Your personal info stays private while using Brave Talk, per our",
   "footer_pst_text": "privacy policy",
   "footer_status_text": "Service status",
+  "free_call_limit": "Up to 4 people. Unlimited with",
+  "free_call_premium_link": "Premium",
```
DJAndries: We should probably have one translation with an embedded link for this, since the placement of the word "Premium" will depend on the language. See `download_extension_footer` for an example of a translation with an embedded link.

src/i18n/locales/en/translation.json
```diff
@@ -19,11 +19,15 @@
   "footer_pre_text": "Your personal info stays private while using Brave Talk, per our",
```
DJAndries: Could we ask the agent to purge any unused translations, if any?

src/components/Recordings.tsx
```diff
@@ -16,21 +15,88 @@ interface Props {
...
+const ExpiryLabel = ({ mobileOnly = false }: { mobileOnly?: boolean }) => {
+  const { t } = useTranslation();
+
+  return (
+    <span
+      css={{
+        display: mobileOnly ? "none" : "inline-flex",
+        flexShrink: 0,
+        padding:
+          "calc(var(--leo-spacing-xs) / 2) calc(var(--leo-spacing-s) + var(--leo-spacing-xs) / 2)",
```
DJAndries: Same question as above regarding this complex css padding property. Is this how the property is specified in figma?

src/jitsi/options.ts
aguscruiz: It's because for us it doesn't really make sense to have a button in the prejoin screen when the clear intention is to connect to a call. In this instance if you don't want to join the call it's easier to just close the tab. Also, we had some reports of people clicking the hang up button by reflex thinking it was the join button, since it's more eye catching than the "join" one

src/i18n/locales/en/translation.json
aguscruiz: No, you're right, that's a bug to fix

src/i18n/locales/en/translation.json
aguscruiz: Fixed

src/i18n/locales/en/translation.json
aguscruiz: Fixed

src/i18n/locales/en/translation.json
aguscruiz: Fixed

src/components/Button.tsx
aguscruiz: Yeah you're right, I don't know why Grok decided to make it so complicated. Figma is way simpler actually, like the one I'm using now

src/components/Recordings.tsx
aguscruiz: Same fix applied

src/i18n/locales/en/translation.json
aguscruiz: Done — `free_call_limit` is now a single key with the Premium link embedded via <Trans> (link is the <2> child, same as `download_extension_footer`). For Japanese the key reads 4人まで。<2>プレミアム</2>なら無制限。, so the link sits where the language needs it. The old standalone `free_call_premium_link` key was renamed to `premium_badge` since it still labels the "Leo AI for meetings" pill. tsc, eslint, prettier, and all 9 tests pass

src/i18n/locales/en/translation.json
aguscruiz: Done! Removed all unused translations

src/i18n/locales/en/translation.json
aguscruiz: fixed

src/i18n/locales/en/translation.json
aguscruiz: No, typo, fixed

src/components/Button.tsx
aguscruiz: Fixed

src/components/Recordings.tsx
aguscruiz: Fixed

src/i18n/locales/en/translation.json
```diff
@@ -19,11 +19,15 @@
...
   "footer_pst_text": "privacy policy",
   "footer_status_text": "Service status",
+  "free_call_limit": "Up to 4 people. Unlimited with <2>Premium</2>.",
+  "premium_badge": "Premium",
```
aguscruiz: Fixed in e3dba55: merged the two keys into a single `free_call_limit` string with an embedded `<Trans>` link (same pattern as `download_extension_footer`). The standalone "Premium" badge now uses a separate `premium_badge` key.

src/i18n/locales/en/translation.json
aguscruiz: Fixed in https://github.com/brave/brave-talk/commit/e3dba55e3f80d03fe37adf5ab602a09009811e4a: merged the two keys into a single free_call_limit string with an embedded <Trans> link (same pattern as download_extension_footer). The standalone "Premium" badge now uses a separate premium_badge key.

src/i18n/locales/en/translation.json
aguscruiz: Purged 89 unused translations (45 en / 44 jp). All removed keys were orphaned by removed features — the web3/NFT/BAT-gating keys (avatar_*, bat_gating_*, moderator_*nft*, wallet_*, sign_request_*, etc.) from PR #1598, plus stale strings from deleted guest-screen code (enter_room_button, join_call, talk_title_premium, pass-through keys like Created meeting room). Verified no source reference (literal, dynamic, or Trans) exists; typecheck and lint pass.

